PR #12, opened 2018-04-04
Implementation of Emotion + Media Queries

nmakuch:
This PR consists of:
- Adding the emotion library and it's dependencies to the project
- Importing and linking the styles.js file to use the theme object
- Media Queries


pcraig3:
Since the other PR #14 contains this same commit, I'm going to close this one and merge that one.

PR #14, opened 2018-04-04
Top banners

nmakuch:
This PR consists of:
- Migrating the Alpha/Federal Banners and Footer components from nrcan project into the new IRCC project
- Importing these components into HomePage before implementing to others

### DIFF
--- a/src/AlphaBanner.js
+++ b/src/AlphaBanner.js
@@ -1,0 +1,59 @@
+import React from 'react'
+import { css } from 'react-emotion'
+import { theme, roundedEdges, mediaQuery } from './styles'
+
+
+const banner = css`
+  background-color: ${theme.colour.grey};
+  color: ${theme.colour.white};
+  padding-left: ${theme.spacing.xxxl}px;
+  padding-right: ${theme.spacing.xxxl}px;
+  padding-top: 12px;
+  padding-bottom: ${theme.spacing.sm}px;
+`
+
+const alpha = css`
+  font-size: ${theme.font.xs};
+  margin-bottom: 0;
+`
+
+const statusBar = css`
+  text-transform: uppercase;
+  line-height: 2;
+  color: white;
+  position: relative;
+  display: inline-block;
+  font-size: ${theme.font.xs};
+  bottom: 2px;
+  padding: 2px ${theme.spacing.md}px;
+  margin-right: ${theme.spacing.md}px;
+  background-color: ${theme.colour.blueAlpha};
+  ${roundedEdges};
+  ${mediaQuery.small(css`
+    font-size: 10px;
+  `)};
+`
+
+const link = css`
+  color: ${theme.colour.white};
+  padding: 0;
+`
+
+const AlphaBanner = () => (
+  <div className={banner}>
+    <p className={alpha}>
+      <span className={statusBar}>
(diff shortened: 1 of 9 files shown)


src/AlphaBanner.js
```diff
@@ -0,0 +1,59 @@
...
+        className={link}
+        href="mailto:paul.craig@cds-snc.ca?subject=Hey,%20I%20love%20your%20app!"
+      >
+        To provide feedback, email us
+      </a>.
+    </p>
+  </div>
+)
+
+export default AlphaBanner
```
pcraig3: I think we should be including tests for these components now that we have jest set up. 
Doesn't have to be too involved since this particular component doesn't take any params -- [something like what we were doing in the nrcan_poc would be fine](https://github.com/cds-snc/nrcan_poc/blob/master/components/__tests__/AlphaBanner.test.js).

src/FederalBanner.js
```diff
@@ -0,0 +1,31 @@
...
+        width="250px"
+        height="24px"
+        flag={theme.colour.white}
+        text={theme.colour.white}
+      />
+    </div>
+  </section>
+)
+
+export default FederalBanner
```
pcraig3: Same as above -> add a test file.

src/Footer.js
```diff
@@ -0,0 +1,34 @@
...
+    <WordMark
+      width="134px"
+      height="32px"
+      flag={theme.colour.white}
+      text={theme.colour.white}
+    />
+  </footer>
+)
+
+export default Footer
```
pcraig3: Same as above -> add a test file.

src/server.js
```diff
@@ -41,6 +42,7 @@ server
       })
       res.send(html)
     } catch (error) {
+      console.log(error.message)
```
pcraig3: Oh yikes. Let's get rid of this.

src/styles.js
```diff
@@ -0,0 +1,101 @@
+import { css } from 'react-emotion'
+
+export const fontFace = css`
+  @import url(https://fontlibrary.org/face/hk-grotesk);
```
pcraig3: As an after-Monday improvement, we should figure out how to host this ourselves I think (or at least discuss options).

src/styles.js
```diff
@@ -0,0 +1,101 @@
...
+    blueAlpha: '#4A90E2',
+    red: '#A5071B',
+    grey: '#4A4A4A',
+    gray: '#4A4A4A',
+    greyLight: '#DBDBDB',
+    grayLight: '#DBDBDB',
+    white: '#FFFFFF',
+    black: '#000000',
+    focus: '#ffbf47',
+    visited: '#551a8b',
```
pcraig3: We don't need most of these colours anymore, so we should clear out the ones we're not using. Probably this means getting in touch with Zack.

src/HomePage.js
```diff
@@ -1,10 +1,36 @@
...
+import FederalBanner from './FederalBanner'
+import Footer from './Footer'
+
+injectGlobal`
+
+html, body {
+    padding: 0;
+		margin: 0;
+		background: ${theme.colour.white};
+		height: 100%;
```
pcraig3: Indentation here 😬

PR #18, opened 2018-04-05
Pagestyling

nmakuch:
This PR consists of:
- Refining the styles.js component
- Adding the PageHeader component
- modifying Headers + Footer


src/AlphaBanner.js
```diff
@@ -2,14 +2,13 @@ import React from 'react'
...
-  background-color: ${theme.colour.grey};
+  background-color: ${theme.colour.black};
   color: ${theme.colour.white};
-  padding-left: ${theme.spacing.xxxl}px;
-  padding-right: ${theme.spacing.xxxl}px;
+  padding-left: ${theme.spacing.xxxl};
+  padding-right: ${theme.spacing.xxxl};
   padding-top: 12px;
-  padding-bottom: ${theme.spacing.sm}px;
+  padding-bottom: ${theme.spacing.sm};
```
pcraig3: Yeah, this is a good call. remembering to call `px` every time was kind of annoying.

src/Footer.js
```diff
@@ -4,31 +4,39 @@ import styled, { css } from 'react-emotion'
...
-    </Link>
-    <WordMark
-      width="134px"
-      height="32px"
-      flag={theme.colour.white}
-      text={theme.colour.white}
-    />
-  </footer>
+  <section>
+    <div className={topBar} />
```
pcraig3: I think this should be configurable. Since we have a version of the footer that doesn't need a top bar, we should make this optional. (so, for example, you can pass in a property that decides whether or not our footer will have a top bar:  `<Footer topBar={true} />` or something)

src/HomePage.js
```diff
@@ -14,29 +15,42 @@ html, body {
 		margin: 0;
 		background: ${theme.colour.white};
 		height: 100%;
```
pcraig3: Hey d00d, what is up with this weird indentation???????

src/HomePage.js
```diff
@@ -14,29 +15,42 @@ html, body {
...
         <FederalBanner />
         <main role="main">
-          <section>
-            <header>Reschedule your Canadian Citizenship appointment</header>
-          </section>
-          <section>
-            <h1>Register</h1>
+          <PageHeader>
+            <h1 className="title">
+              Reschedule your Canadian <br /> Citizenship appointment
```
pcraig3: I think it's a better idea to set a percentage width on the `<h1>` component rather than enforce a line break.

src/HomePage.js
```diff
@@ -14,29 +15,42 @@ html, body {
...
     `)};
-
 	}
 
+.title {
+    font-size: ${theme.font.lg2};
+    font-family: ${theme.weight.b};
+}
+
+.content {
```
pcraig3: We could make this something like `main > section` or something. Would that work?

Then we could avoid having classnames in our code?

src/styles.js
```diff
@@ -55,33 +55,35 @@ export const theme = {
...
+    md: '1em',
+    sm: '0.5em',
+    xs: '0.33em',
+    xxs: '0.17em',
+    lg: '1.3em',
+    xl: '2em',
+    xxl: '3em',
+    xxxl: '4em',
+
+    lgXl: '2.75em',
```
pcraig3: What is `lgXl`??????

Just redo the other ones so that we don't break our paradigm.

src/Footer.js
```diff
@@ -4,31 +4,42 @@ import styled, { css } from 'react-emotion'
...
+  {
+    height: '0.4em',
+    border: 'none',
+  },
+  props => ({ background: props.background }),
+)
+
+const Footer = ({ hasTopBar, backGround }) => (
+  <section>
+    {hasTopBar ? <TopBar background={backGround} /> : ''}
```
pcraig3: If we made these both into one variable, we could (i) check if it exists, and then (ii) pass it into the topBar in one go.

PR #23, opened 2018-04-06
Page-wide styling + tests

nmakuch:
This PR consists of:
- Creating tests for new components
- Applying styling to all other pages

### DIFF
--- a/src/Footer.js
+++ b/src/Footer.js
@@ -4,6 +4,14 @@ import { WordMark } from '@cdssnc/gcui'
 import styled, { css } from 'react-emotion'
 import { theme } from './styles'
 
+const Circle = styled.span`
+  font-size: 8px;
+  padding-left: ${theme.spacing.lg};
+  padding-right: ${theme.spacing.lg};
+  position: relative;
+  bottom: 2px;
+`
+
 const footer = css`
   background-color: ${theme.colour.white};
   padding: ${theme.spacing.xl} ${theme.spacing.xxxl};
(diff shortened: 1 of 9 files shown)


src/__tests__/Footer.test.js
```diff
@@ -0,0 +1,15 @@
+import { shallow } from 'enzyme'
+import Footer from '../Footer'
+import React from 'react'
+
+describe('<Footer />', () => {
+  it('renders footer', () => {
+    const footer = shallow(<Footer />)
+    expect(footer.find('footer')).toBeTruthy()
+  })
```
pcraig3: I think it is also worth checking that there is no `hr` element in here.
`expect(footer.find('hr')).toBeFalsey()`

src/__tests__/Footer.test.js
```diff
@@ -0,0 +1,15 @@
...
+
+  it('renders footer with topBar', () => {
+    const topBar = shallow(<Footer topBarBackground="black" />)
+    expect(topBar).toMatchObject(topBar)
```
pcraig3: Here we are just checking that the object returned on line 12 is the same as itself.
Instead, we should be checking that there is a `footer` element, as well as an `hr` element.
expect(footer.find('footer')).toBeTruthy()
expect(footer.find('hr')).toBeTruthy()

src/Footer.js
```diff
@@ -30,9 +38,20 @@ const Footer = ({ topBarBackground }) => (
...
+      <div>
+        <Link href="https://www.canada.ca/en/transparency/privacy.html">
+          Privacy
+        </Link>
+
+        <Circle> &#9679; </Circle>
+
+        <Link href="#">Terms and Conditions</Link>
+
+        <Circle> &#9679; </Circle>
```
pcraig3: Oh, weird. That's cool.

PR #25, opened 2018-04-09
Added buttons to corresponding navlinks - fits the wireframe

Amirroracle:
Please make sure that the button tests follow standard practice and such.


src/__tests__/Button.test.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react'
+import { shallow } from 'enzyme'
+import Button from '../forms/Button'
+
+describe('<Button>', () => {
+  it('Renders and tests disable function', () => {
+    const buttonDisabled = shallow(<Button disabled={true}>Next</Button>)
+    expect(buttonDisabled.props().disabled).toBeTruthy()
+
+    const buttonDefault = shallow(<Button >Next</Button>)
```
pcraig3: Ooo, this should a be a second test. So create a new `it('does something') => { // test }` for testing different button states rather than putting both into one test.

src/forms/Button.js
```diff
@@ -8,7 +8,7 @@ const button = css`
   font-size: ${theme.font.md};
   font-family: ${theme.weight.s};
   color: ${theme.colour.white};
-  background-color: ${theme.colour.blue};
+  background-color: ${theme.colour.gray};
```
pcraig3: I've just had a look and it's still blue in some states, because there is a new colour set on buttons when they are active or focused rather than just a change in opacity.
https://github.com/cds-snc/ircc-rescheduler/blob/985db159eb9ed2a99c3213bb9cf5d097a2ae9769/src/forms/Button.js#L27-L31

PR #26, opened 2018-04-09
Add react-day-picker

sleepycat:
:tada:


src/Calendar.js
```diff
@@ -0,0 +1,254 @@
...
+      <div>
+        <DayPicker
+          numberOfMonths={2}
+          disabledDays={[{ daysOfWeek: [0, 6] }]}
+          pagedNavigation
+          fixedWeeks
+          selectedDays={this.state.selectedDays}
+          onDayClick={this.handleDayClick}
+        />
+      </div>
```
pcraig3: Does this need to be wrapped in a `<div>`? Is it better for us adding styling to it or anything? Otherwise we could just return the `DayPicker` directly.

src/CalendarPage.js
```diff
@@ -28,7 +29,6 @@ class CalendarPage extends React.Component {
               </Bold>{' '}
               so we can schedule your new appointment.
             </p>
-            {/* what is "help"??? */}
```
pcraig3: lol, I think that was me.

src/Calendar.js
sleepycat: :thinking: True. I'll remove it.

pcraig3:
👍

PR #38, opened 2018-04-12
Separate the text input from text area

Amirroracle:
text area is now a seperate module from text input. The TextAreaAdapter is set to accept an input var which contains properties like name,id, etc..

### DIFF
--- a/src/forms/TextInput.js
+++ b/src/forms/TextInput.js
@@ -19,7 +19,13 @@ const text_input = css`
     width: 100%;
   `)};
 `
-
+const text_area = css`
+  height: 10em;
+  resize: none;
+  width: 550px;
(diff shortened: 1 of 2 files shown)


src/forms/TextInput.js
```diff
@@ -19,7 +19,13 @@ const text_input = css`
     width: 100%;
   `)};
 `
-
+const text_area = css`
+  height: 10em;
+  resize: none;
+  width: 550px;
+  margin-top: ${theme.spacing.sm};
+  ${text_input};
```
pcraig3: If you move this include to the top (ie, to line 23), then the following CSS will be later in precedence. This means that none of the CSS rules you have on lines 23-26 risks being overwritten by some styling you're pulling in inside of this variable.

PR #42, opened 2018-04-16
404 page setup

nmakuch:
This PR consists of:
- Mock 404 page with placeholder text


src/404Page.js
```diff
@@ -0,0 +1,45 @@
...
+          <PageHeader>
+            <H1>Request a new Canadian Citizenship test date</H1>
+          </PageHeader>
+
+          <ErrorContent>
+            <H1> 404: Page not found </H1>
+            <H2>
+              {' '}
+              Oops! We cant seem to find the page you are looking for:
+            </H2>{' '}
```
pcraig3: I don't think we need these `{' '}` for anything.

PR #43, opened 2018-04-16
Create Layout component, reorder some CSS 

pcraig3:
Our pages all have common headers and footers, so some kind of Layout abstraction will help us maintain that consistency.

Also helps us move some of that global CSS, which means that we can move away from creating a ton of global styling on the Home page.

**Note**

I didn't put this is as a post-it note earlier today because I forgot about it. But I had been working on this since Friday (ie, I was already halfway through it), so I figured I should push on.


pcraig3:
Thanks @nmakuch!

PR #46, opened 2018-04-18
Mobile styling

nmakuch:
This PR consists of:
-Media Queries for small devices

### DIFF
--- a/src/AlphaBanner.js
+++ b/src/AlphaBanner.js
@@ -10,14 +10,19 @@ const banner = css`
   padding-right: ${theme.spacing.xxxl};
   padding-top: 12px;
   padding-bottom: ${theme.spacing.sm};
+
+  ${mediaQuery.small(css`
+    padding-left: ${theme.spacing.xl};
+    padding-right: ${theme.spacing.xl};
+  `)};
 `
 
 const alpha = css`
-  font-size: ${theme.font.sm};
+  font-size: 0.75rem;
   margin-bottom: 0;
 
   ${mediaQuery.small(css`
-    font-size: 8px;
+    font-size: ${theme.font.xs};
   `)};
 `
 
@@ -34,7 +39,7 @@ const statusBar = css`
   background-color: ${theme.colour.white};
   ${roundedEdges};
   ${mediaQuery.small(css`
-    font-size: 8px;
+    font-size: ${theme.font.xs};
     padding: 0px ${theme.spacing.md};
   `)};
 `
--- a/src/AlphaBanner.js
+++ b/src/AlphaBanner.js
@@ -13,8 +13,12 @@ const banner = css`
 `
 
 const alpha = css`
-  font-size: ${theme.font.xs};
+  font-size: ${theme.font.sm};
   margin-bottom: 0;
+
+  ${mediaQuery.small(css`
+    font-size: 8px;
+  `)};
 `
 
 const statusBar = css`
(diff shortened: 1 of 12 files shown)


src/forms/TextInput.js
```diff
@@ -8,13 +8,17 @@ const text_input = css`
...
     outline-offset: 0px;
   }
 
+  ${mediaQuery.small(css`
+    width: 100%;
+  `)};
+
   ${mediaQuery.xs(css`
     width: 100%;
   `)};
```
pcraig3: We can get rid or this `xs` rule since the small rule will cover it.

src/styles.js
```diff
@@ -99,7 +99,7 @@ export const theme = {
   font: {
     xxs: '0.2em',
     xs: '0.5em',
-    sm: '0.75em',
+    sm: '0.9em',
     base: '1em',
     md: '1em',
     lg: '1.2em',
```
pcraig3: I think our `xs` and `xxs` font sizes are too low in general. If `0.5em` is too small to be readable, we should just move our scale down. 

Maybe we delete `xxs` and we make `xs` larger.

src/styles.js
```diff
@@ -191,10 +192,23 @@ export const CalHeader = styled.div`
...
   padding: ${theme.spacing.xl} 0 ${theme.spacing.lg} 0;
 `
+
+export const Cancel = styled.p`
+  display: inline;
+
+  ${mediaQuery.xs(css`
+    margin-top: ${theme.spacing.md};
+    display: block;
+  `)};
```
pcraig3: Ugh, I hate `styled.p`s. You can keep this in if you want, but I'm coming for this later.

src/AlphaBanner.js
```diff
@@ -13,8 +13,12 @@ const banner = css`
 `
 
 const alpha = css`
-  font-size: ${theme.font.xs};
+  font-size: ${theme.font.sm};
   margin-bottom: 0;
+
+  ${mediaQuery.small(css`
+    font-size: 8px;
+  `)};
```
pcraig3: No way, we should use `xs` or something.

src/AlphaBanner.js
```diff
@@ -30,7 +34,8 @@ const statusBar = css`
   background-color: ${theme.colour.white};
   ${roundedEdges};
   ${mediaQuery.small(css`
-    font-size: 10px;
+    font-size: 8px;
```
pcraig3: `xs` 😬

src/styles.js
pcraig3: Also, could we make these into `rem`s instead of `em`s? Please please please? `rem`'s are [pretty well supported](https://caniuse.com/#search=rem).

pcraig3:
## mobile screenshots comparison

| NRCAN | IRCC  |
|--------|-------|
| alpha badge and federal header is large enough to kind of read | alpha badge and federal header are very hard to read |
|  20180418_215155000_ios  |  20180418_214343000_ios |
| body text is pretty large here -- headers are emphasized by font weight more than size | body text is pretty small compared to headers |
|   20180418_214306000_ios  |  20180418_215234000_ios  |
| form labels are submit button are huge and easy to click | let's make everything bigger pls. Also the footer links aren't centred |
|  20180418_214221000_ios  |  20180418_214328000_ios |

zakkain:
image

at the iPhone 5/5s size, there is extra padding on the left side that introduces horizontal scrolling on the page - which when scrolled all the way left, cuts off the right-hand side of the form

pcraig3:
Killing this since #53 went in.

PR #53, opened 2018-04-20
Mobile/desktop whatever

pcraig3:
Doesn't matter where we are viewing this, we always expect a handsome interface. 

Works great on my phone.


pcraig3:
Gonna merge and make this our live version.

PR #57, opened 2018-04-20
Form data persistence

Amirroracle:
-make the default values for the homepage schema in gql to be empty strings 
-initialize form with default values, if user goes back the form values are initialized
-Form init no longer throws an error in the try catch since the gql query is initialized when mounted.
-state contains form values, and passed the values to the initialValues prop in the Form component.
-add a 'checked' prop to the radiobuttons, allow itself to be initialized.


Amirroracle:
Looking at the test that is failing - also tangentially fixing a routing issue

PR #60, opened 2018-04-23
Added ReviewPage

nmakuch:
This PR consists of:
- Creating the static version of the Review Page.
- Creating the corresponding routes to this page.
- Added some styles to styles.js that are being used on multiple pages
- Media Queries for the Review Page

### DIFF
--- a/web/src/LandingPage.js
+++ b/web/src/LandingPage.js
@@ -10,6 +10,10 @@ const contentClass = css`
   p {
     padding-bottom: ${theme.spacing.lg};
   }
+
+  li {
+    padding-bottom: ${theme.spacing.sm};
+  }
 `
 
 const list = css`
@@ -29,11 +33,11 @@ const landingSection = css`
 `
 
 const moreInfo = css`
-  margin-left: ${theme.spacing.lg};
-  padding-bottom: 0rem;
+  font-family: ${theme.weight.l};
+  font-size: ${theme.font.md};
 `
 
-class RegistrationPage extends React.Component {
+class LandingPage extends React.Component {
   render() {
     return (
       <Layout contentClass={contentClass}>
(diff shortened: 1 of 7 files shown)


web/src/LandingPage.js
```diff
@@ -48,31 +52,36 @@ class RegistrationPage extends React.Component {
           <H2>You will need:</H2>
           <ul className={list}>
             <li>
-              your paper file number (found at the top of your test notice
-              email)
+              your paper file number<br />
+              <span className={moreInfo}>
+                This is found at the top of your test notice email.
+              </span>
```
pcraig3: Hmm, I think it would be better to make these a bit more structured instead of using a `<br />`

You can put whatever you want inside `<li>` elements, so we can stick more markup in there.

It looks like we have two elements: first one is kind of like a heading and the second one is (optionally) some details.
Maybe we could do 
```
<li>
<p>your paper file number</p>
<span>this is found at the top of your test notice email.</span>
</li>
```

Or you could make it like `<li><div>heading</div><p>details</p></li>`, if that looks better to you.

Kind of open to whatever, just think we want to be a bit more specific.

web/src/ReviewPage.js
```diff
@@ -0,0 +1,204 @@
...
+              <Trans>Cancel</Trans>
+            </NavLink>
+          </div>
+        </BottomContainer>
+      </Layout>
+    )
+  }
+}
+
+export default ReviewPage
```
pcraig3: So this page actually looks really awesome 👏 👏, but I have a couple of feedbacks.

### 1. responsive tables

tables are bad for layouts because they are hard to make responsive. But tables are great for screenreaders, because you are very clearly structuring how your data relates to each other. 
ie, in our case, we have a table-like structure where
- column 1 is things we are asking you for
- column 2 is things you have answered
- column 3 is a button/link, allowing you to update your answer

So that would be better to use a real table, I think, in terms of describing our data. 
**But** tables have terrible responsive solutions. So, I don't think we want to move backwards on this.

### 2. too many different divs

I think it would be better to be a bit stricter about our elements here.
If we're using divs for the table, we should limit ourselves to 4.

- `Row` div (with a dotted bottom border)
- `Col1` div for all form headings
- `Col2` div for all answers
- `Col3` div for all buttons/links

If we are eventually porting this over to a table instead of divs, this will make it easier for us.

PR #61, opened 2018-04-23
Revert CircleCi config

pcraig3:
Back to how it used to be when we had green checkmarks. 

If the branch is called `add-deploy`, it will run through the whole deploy process for us so we don't have to merge this change to master to see how it will turn out.


Amirroracle:
looks good

PR #72, opened 2018-04-25
Ecs deploy

szinck1:
Deploy PR will deploy `ircc-rescheduler` to AWS at the url https://ircc-web.cdssandbox.xyz.


sleepycat:
We need to get the other containers (API and Traefik) folded into this deploy script before merging. Traefik looks like it will be useful for tackling both [rate limiting](https://docs.traefik.io/configuration/commons/#rate-limiting) and [HTTPS](https://docs.traefik.io/user-guide/docker-and-lets-encrypt/) which are both on our todo list.

pcraig3:
@szinck1 is off this project and it looks like we're not planning to deploy to ECS anymore. Shutting this down since we're not planning to do any more work on it.

PR #90, opened 2018-05-03
Add makefile with docker-compose dev file

sleepycat:
Start the app in development mode with `make dev`.
If jest is installed globally you can also run `make test`.

Minor change to web service to get webpack-dev-server listening on all ports
rather than just locally so that it works in a container.

You will notice that port 3005 is mapped directly to the host rather than going
through the proxy like 3004. This is a workaround for fact that I couldn't
figure out how to get Traefik to properly route to both 3004 & 3005. I will
fight that battle again another day.

Also note the environmental variables that are expected to exist.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,5 @@
+dev:
+	docker-compose -f dev.yml -p dev up
+test:
+	jest --projects web/ api/
+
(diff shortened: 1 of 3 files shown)


pcraig3:
Cool, we'll go over this in more detail at the Docker meetup later, right?

PR #93, opened 2018-05-04
Simplify Button component

sleepycat:
Simply export the component created by styled.button.

### DIFF
--- a/web/src/forms/Button.js
+++ b/web/src/forms/Button.js
@@ -1,9 +1,7 @@
-import React from 'react'
-import PropTypes from 'prop-types'
-import { css } from 'react-emotion'
+import styled, { css } from 'react-emotion'
 import { theme, roundedEdges, mediaQuery, incrementColor } from '../styles'
 
-const button = css`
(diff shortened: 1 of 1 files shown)


web/src/forms/Button.js
```diff
@@ -44,16 +42,4 @@ const button = css`
   }
 `
 
-const Button = ({ disabled = false, children }) => (
```
pcraig3: I think it would be cool to keep the `children` validation (ie, we will only allow one element as a child), although we can get rid of everything else.

Something like:
```
export default ({ children, ...props })  => (
  <Button {...props} >
    {children}
  </Button>
)

/* validation to make sure only one child is passed in */	
Button.propTypes = {	
  children: PropTypes.element.isRequired,	
}
```

Or do you think that's not worth keeping?

sleepycat: What we want is simply the default behaviour of React, so no real need for any of this. Stating that children are required really would only stop people from doing `<Button />` which doesn't even make sense.

Issue #78, opened 2018-04-27
Clean up HTML Markup


Amirroracle:
Currently there are many Html warnings that need fixing, though some like the following

> The type attribute is unnecessary for JavaScript resources.

can be ignored, but others are more important. This was run only on the homepage, did not include issues from the other pages.

[https://validator.w3.org/nu/?showimagereport=yes&doc=http%3A%2F%2Frescheduler.cds-snc.ca%2F](url)

<img width="1292" alt="screen shot 2018-04-27 at 12 22 40 pm" src="https://user-images.githubusercontent.com/9355166/39373622-23eb1514-4a16-11e8-92a2-155ed1217c77.png">
<img width="969" alt="screen shot 2018-04-27 at 12 26 37 pm" src="https://user-images.githubusercontent.com/9355166/39373653-41774616-4a16-11e8-8c40-edc083572003.png">

pcraig3:
Some of these warnings are things we want to keep around, or they're being automatically generated for us. 
If we're misconfiguring things we should fix that, but if we're doing everything right and there's some library generating a warning, then we don't need to spend too much time on those.

As an example, the `main role="main"` warning is something we are disregarding. [It's not needed for modern browsers, but it's useful for Internet Explorers](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/main#Browser_compatibility).

pcraig3:
Just looked on the calendar page to see what the specific problems are:

## Errors

### 1. Element title not allowed as child of element head in this context
### 2.  Element title must not be empty.

Our page titles aren't really figured out. It would be great to have a server-rendered page-title thing for each page. Otherwise, we're always going to end up with two `<title>`s in the `<head>`.

### 3. CSS: transform: is an incorrect operator

Don't know where this is coming from, but if this is our code is should be easy to fix.

## Warnings 

### 1. Consider adding a lang attribute to the html start tag to declare the language of this document

The `lang` attribute tells screenreaders what language to expect the body text to be in. If it changes while they are on the page, then the screenreader may not know that this has changed. We may need to force a full-page reload when the language is changed to get around this.

But even if we don't we need to be able to change the attribute back and forth as users switch between english and french. 

### 3. The charset attribute on the script element is obsolete.
### 4. The type attribute is unnecessary for JavaScript resources.

This is being generated outside of our code and they're just warnings so we don't care.

### 5. Section lacks heading

[...]

PR #96, opened 2018-05-04
Html validation fixes - fixed 4 of them 

Amirroracle:
-Fix html error regarding a style within a span - also add the ircc website
-remove main role
-add title fix but still need to add translation
- move href class to the <a> tag

### DIFF
--- a/web/src/Layout.js
+++ b/web/src/Layout.js
@@ -57,7 +57,7 @@ const Layout = ({ children, contentClass = '' }) => (
       </span>
     </AlphaBanner>
     <FederalBanner />
-    <main role="main">
+    <main>
       <PageHeader>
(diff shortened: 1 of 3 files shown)


web/src/Layout.js
```diff
@@ -57,7 +57,7 @@ const Layout = ({ children, contentClass = '' }) => (
       </span>
     </AlphaBanner>
     <FederalBanner />
-    <main role="main">
```
pcraig3: I think we should keep this in. I wrote about this previously here: https://github.com/cds-snc/ircc-rescheduler/issues/78#issuecomment-385426486

pcraig3:
I think we should keep role="main" but then merge away!

PR #97, opened 2018-05-04
Change edit links and de-clutter summary page

Amirroracle:
-Moved some of the css to the styles.js, should perhaps start thinking of indexing/de-cluttering the styles.js page since it is beginning to get a tad large.
-Set up a better abstraction for the table in the summary page. A lot less clutter.
-Added the links according to the design doc for the 'Edit' and 'Change' links in the summary page. 

Todo:
-wait until @pcraig3's merge for the calendar to go through and add the calendar data to the availability row. 
-add tests for the new components I added.


web/src/Summary.js
```diff
@@ -73,97 +39,27 @@ export const Summary = ({
...
-        )}
-      </Column2>
-      <Column3>
-        <TextLink className={change} href="#">
-          <Trans>Change</Trans>
-        </TextLink>
-      </Column3>
-    </Row>
+    <UserInfoRow
+      header="Full name"
```
pcraig3: I think it's better if these are `<Trans>` elements, so we can change the name between English and French. 

ie
```
header={<Trans>Full name</Trans>} 
```

And then you could update the proptypes at the bottom to be 
`header: PropTypes.element.isRequired`.

web/src/Summary.js
```diff
@@ -73,97 +39,27 @@ export const Summary = ({
...
+      secondColumn={fullName}
+      thirdColumn="Edit"
```
pcraig3: In the new mockup, the text for the third column is always the same. 
The link changes, but the link text is the same.

image

web/src/TableContents.js
```diff
@@ -0,0 +1,52 @@
...
+      </H2>
+    </Column1>
+    <Column2>{secondColumn}</Column2>
+    <Column3>
+      <TextLink className={change} href="/calendar">
+        <Trans>{thirdColumn}</Trans>
+      </TextLink>
+    </Column3>
+  </Row>
+)
```
pcraig3: There's a lot of similar code for `Column1`, `Column2` (now that the dates are in) and `Column3` across both of these components. You could probably generalise this further. 
(but up to you; I'm not so bothered)

web/src/TableContents.js
```diff
@@ -0,0 +1,52 @@
...
+        <Trans>{header}</Trans>
+      </H2>
+    </Column1>
+    <Column2>
+      <p>{secondColumn}</p>
+    </Column2>
+    <Column3>
+      <TextLink href="/register">
+        <Trans>{thirdColumn}</Trans>
+      </TextLink>
```
pcraig3: We want `NavLinks` here.

If you can delete `TextLink` outright, I would be pleased. But if it's still being used somewhere else, then nvm.

web/src/TableContents.js
```diff
@@ -0,0 +1,52 @@
...
+  header: PropTypes.string,
+  secondColumn: PropTypes.object,
+  thirdColumn: PropTypes.string,
+}
+
+UserInfoRow.propTypes = {
+  header: PropTypes.string,
+  secondColumn: PropTypes.oneOfType([PropTypes.string, PropTypes.element]),
+  thirdColumn: PropTypes.string,
+}
```
pcraig3: Both DateRow and UserInfoRow can accept the same proptypes now.

I think they both take: 
```
{
  header: PropTypes.element.isRequired,
  secondColumn: PropTypes.oneOfType([PropTypes.string, PropTypes.element]).isRequired,   
}
```

And then for the third column, you could pass in a string if you wanted (which would be the href).
ie 
```
  thirdColumn: PropTypes.string.isRequired
```

web/src/Summary.js
Amirroracle: Nvm, I think I found a decent abstraction

web/src/Summary.js
Amirroracle: yup, good point

web/src/Summary.js
```diff
@@ -73,97 +40,51 @@ export const Summary = ({
...
+        </NavLink>
+      }
+    />
+    <SummaryRow
+      header={<Trans>Paper file number</Trans>}
+      secondColumn={paperFileNumber}
+      thirdColumn={
+        <NavLink to="/register">
+          <Trans>Change</Trans>
+        </NavLink>
```
pcraig3: Either this, or your `thirdColumn` could just pass in a string that's the `to` string.

So you could be like 

```
// Summary.js
<SummaryRow
  ...
  thirdColumn="/register"
/>

// TableContents.js
const SummaryRow ...
  <Row>
    <Column3>
      <NavLink to={thirdColumn}>
        <Trans>Change</Trans>
      </NavLink>
    </Column3>
  </Row>
```

Up to you.

web/src/TableContents.js
```diff
@@ -0,0 +1,21 @@
...
+  </Row>
+)
+
+SummaryRow.propTypes = {
+  header: PropTypes.object.isRequired,
+  secondColumn: PropTypes.oneOfType([PropTypes.string, PropTypes.element]),
+  thirdColumn: PropTypes.object.isRequired,
+}
+
+export { SummaryRow }
```
pcraig3: This file is pretty small now and only contains 1 component. 

I would suggest:
1. put this code straight into `Summary.js` and delete this file.
2. rename the file to `SummaryRow.js`

pcraig3:
rebased and ready for liftoff! 🚀

PR #100, opened 2018-05-07
Enforce maximum number of dates

### LINKED ISSUE #87: Add logic to limit the number of dates that can be selected from the calendar 🗓
Currently, there is no limit to the maximum number of dates that a user can choose.

Related:
* #83 imposed a minimum selection of 3 dates before a user is allowed to progress to the `/review` view.

Todo: 
* add logic to limit the _maximum_ selectable dates to 3

Suggested flow:
* For a better UX, I'd suggest preventing the user from being able to select more than 3 dates (as opposed to being able to select more than 3 dates, then throwing an error when the user tries to click the `review` button).
* Before 3 dates are selected, the `review` button can be disabled (with a visual indication, such as being greyed out).
* When 3 dates are selected, the button can be enabled. And message can be displayed letting the user know that if they'd like to select another date, they'd need to deselect an existing one.

Current:
screen shot 2018-05-01 at 17 24 35

pcraig3:
Users must select at least 3 dates, but they cannot select more than 3 dates.

Before they select 3, they can't go to the next page.
Once they have selected 3, they shouldn't be able to select any more.

This pull request will restrict users from being able to select more than 3 dates, which is our current limit. If they want a new day, they have to remove an existing day.

In this pull request:
- [x] stop users from selecting more than 3 days
- [x] some kind of on-screen warning when they try to select more than 3 days
- [x] unselect a day by removing it from the list

This pull request will resolve #87 once it is merged.


web/src/Calendar.js
```diff
@@ -249,24 +268,42 @@ class Calendar extends Component {
...
+      dayLimit,
+    } = this.props
+
+    if (!this.state.selectedDays.length && value && value.length) {
+      await this.setState({ selectedDays: value })
     }
 
     const { selectedDays } = this.state
-    if (selected) {
+
```
Amirroracle: could use a comment like "selected means that the current day has been already selected". Just a one liner to denote that the 'selected' variable is a boolean that checks if the day selected by the user has already been selected.

web/src/Calendar.js
```diff
@@ -249,24 +268,42 @@ class Calendar extends Component {
...
+      // add an error message to the internal state and then return early
+      if (selectedDays.length >= dayLimit) {
+        await this.setState({
+          errorMessage: (
+            <Trans>
+              You have already selected the maximum number of dates!
+            </Trans>
+          ),
+        })
+        return
```
Amirroracle: return statement here is likely not needed, should maybe refactor by removing the return statement and just adding an else after the if(!selected) for the rest of the logic.

web/src/Calendar.js
```diff
@@ -249,24 +268,42 @@ class Calendar extends Component {
...
+      }
+
+      // push new day into array of selectedDays
+      selectedDays.push(day)
```
Amirroracle: rather pedantic - but perhaps doing an if condition on the day variable (doesn't seem to be a check for it in this scope) before pushing it to a data structure is good defensive practice. Not totally necessary but I generally like being defensive programming philosophy.

web/src/Calendar.js
```diff
@@ -299,24 +336,41 @@ class Calendar extends Component {
...
+                {value.map((day, index) => (
+                  <li key={index}>
+                    {`${index + 1}. `}
+                    <Time date={day} />{' '}
+                    <button
+                      type="button"
+                      onClick={this.removeDayOnClickOrKeyPress(day)}
+                      onKeyPress={this.removeDayOnClickOrKeyPress(day)}
+                    >
+                      Remove date
```
Amirroracle: Needs translation tag

web/src/Calendar.js
```diff
@@ -299,24 +336,41 @@ class Calendar extends Component {
...
+                    </button>
+                  </li>
+                ))}
+              </ol>
+            ) : (
+              'No dates selected'
```
Amirroracle: Same with here I guess?

web/src/__tests__/Calendar.test.js
```diff
@@ -2,16 +2,35 @@ import React from 'react'
...
+    .at(index)
     .simulate('click')
 }
 
-const defaultProps = ({ value = '' } = {}) => {
+const clickFirstDate = wrapper => {
+  return clickDate(wrapper, 0)
+}
+
+const getDateStrings = wrapper => {
```
Amirroracle: As discussed, this can be refactored to return a map + join call, makes it much easier to read and achieves the same result.

web/src/Calendar.js
pcraig3: This return statement short-circuits the logic. Otherwise, the method would continue by pushing the day into the `selectedDays` array (on line 297).
This is similar to returning early on line 239 if a disabled day is clicked.

There's nowhere else for this logic to go, I don't think. 

First we have to see whether this is a day that is not currently selected -- ie, this happens on line 282 -- and then we have to check if we've hit the maximum day limit.

web/src/Calendar.js
pcraig3: What do you mean? 

- on line 243, we make sure it's in a standard Date format, and 
- on line 282, we check if it's previously selected

web/src/Calendar.js
pcraig3: Yup, good point.

web/src/Calendar.js
Amirroracle: yeah I was wondering if we could do something like remove the return and do something like
`if(!selected){
//some logic then return
}else{ the rest of the function} 
`

it's no big deal but I figure it's easier to debug an if else than a short-circuit

web/src/Calendar.js
Amirroracle: ah but in 243 you are assigning it, what comes from that assignment may not be deterministic. 
I suppose if it is selected it is technically impossible for the day to be null, so you have a point :).

web/src/Calendar.js
pcraig3: The problem with our algorithm is that there's a nested `if` check needed.

```
  --> if this day has not been selected
    ----> if we are already at the max days
            // error message
            // return
    ----> else we are not at the max days
            // add a new day
  --> else this day has previously been selected
        // unselect this day

// do some more stuff
```

I'm not sure how to more cleanly refactor this so I'm inclined to leave it be.

web/src/Calendar.js
pcraig3: I think it's okay.

[`onDayClick`](http://react-day-picker.js.org/api/DayPicker#onDayClick) always hands us back a `Date` object as `day`, and our [`makeGMTDate`](https://github.com/cds-snc/ircc-rescheduler/blob/master/web/src/Time.js#L8) function always returns a date whether it gets passed a string or a Date object.

PR #101, opened 2018-05-08
Fix SVG size, make fonts for text-based input elements sans-serif

pcraig3:
Between Firefox and IE, we were getting a few CSS inconsistencies.

## 1. SVGs were the wrong size

| ie11 | firefox |
|------|---------|
|  image    |   screen shot 2018-05-08 at 2 55 58 pm |

**Solution**
Fixed sizes for svgs. We don't really change their size anyway.

## 2. Fonts in input elements were inconsistent

| ie11 | firefox |
|------|---------|
|  image  |  image |

**Solution**
Explicit font-family rules for text inputs and textareas.

### DIFF
--- a/web/src/FederalBanner.js
+++ b/web/src/FederalBanner.js
@@ -18,6 +18,12 @@ const container = css`
   display: -moz-box;
   display: flex;
 
+  .svg-container {
+    /* same as the width value in our svg */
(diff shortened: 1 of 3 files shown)


pcraig3:
Gonna merge this since the changes are really minor.

PR #104, opened 2018-05-08
Submit dates along with other user data

sleepycat:
This commit adds:
* availability to the RescheduleForm input type
* Validation that availability is a proper date
* Validation that 3 dates are the only acceptable number of dates
* updated query in the web service to include dates

### DIFF
--- a/web/src/queries.js
+++ b/web/src/queries.js
@@ -28,6 +28,7 @@ export const SUBMIT = gql`
     $explanation: String!
     $reason: String!
     $paperFileNumber: String!
+    $availability: [Date!]!
   ) {
     decline(
       input: {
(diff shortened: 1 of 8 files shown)


pcraig3:
Awesome! Got it to work.

web/src/queries.js
```diff
@@ -28,13 +28,15 @@ export const SUBMIT = gql`
     $explanation: String!
     $reason: String!
     $paperFileNumber: String!
+    $availability: [Date!]!
```
pcraig3: This means there must be an array of at least 1 date?

sleepycat: This means availability is a required array, and the array is required to have Dates in it.

PR #107, opened 2018-05-10
Avoid JSDOM test environment for the API

sleepycat:
By default, Jest helpfully creates a document object to help testing of UI
components. For a strictly node.js project like the api, that work is unneeded and
can be skipped to save some time. This change reduces the time to run the test
suite from 1.437s to 0.834s.


sleepycat:
Little victories. :smile_cat:

PR #108, opened 2018-05-10
Add error page

sleepycat:
This adds an error page and logic to display it if the result of the mutation is an error. Tests and refactoring to the next PR as discussed.


sleepycat:
It's working! You are seeing the error page you because the API returned an errors array with a validation error. It looks like you didn't pick three dates. The submission component should help handle this case.

PR #110, opened 2018-05-10
Remove obsolete translations

sleepycat:
As advertised. Just `yarn extract --clean` to remove the kruft.


pcraig3:
Haha, I just did this myself in #100. But yours is isolated, so maybe it's cleaner this way.

Issue #92, opened 2018-05-03
Improve form validation errors on registration page


pcraig3:
**Note** 
Some work has been done on this. [See final comment](https://github.com/cds-snc/ircc-rescheduler/issues/92#issuecomment-391388702).

***

# Problems

There are several usability problems that have been highlighted over a couple rounds of user research concerning our registration form -- specifically around how errors are being shown (or not shown) to users.

## 1. users don't see inline validation errors

Our inline validation errors (that is, errors that appear immediately after the form input loses focus) are being missed by users. The proposed fix for this is to hold off on field validation until the user clicks "Next →".

**Solution: Wait until the user clicks the submit button to validate their inputs.** 

## 2. submit button is disabled unless all fields are valid

Several users have ended up trying to click the submit button but not being able to because of an error further up the page. When errors show up, they are sometimes not aware of them and then confused why they can't click the  button.

**Solution: The button should always be enabled. Users can click it at any point.**

## 3. validation errors aren't being highlighted to users

Currently, when errors appear on the page we are not making users aware of them. This is particularly the case with non-visual interfaces (such as screenreaders). 
A good solution to this is the following:
- validate errors when users click the submit button
- if errors are found, create a component at the top of the page that lists all errors ([here is an example](http://alphagov.github.io/digitalmarketplace-frontend-toolkit/forms/validation.html))
  - error-list component should have a generic header
  - error-list component should list fields with errors and then anchor link to them (ie `<a href="#fullName">`)
  - error-list component should be programatically focused as soon as it appears

An example of this interaction pattern can be found [on the Digital Marketplace](https://www.digitalmarketplace.service.gov.uk/suppliers/create/duns-number). To test it out, enter something in the field that isn't a number and then click "Save and continue".

**Solution: create an error=list component that receives focus, and then lists and links to errors on the page.**

## 4. `<header>` elements are not allowed inside `<label>` elements

[`<label>` elements](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label) can only accept [phrasing content](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Content_categories#Phrasing_content). So we should change the markup for our form labels on the registration page. I would suggest a `<span>` with an appropriate classname. I get that it's not very component-y, but it will solve this problem for us and then we can revisit it later.

# Progress

- [x] users don't see inline validation errors
- [x] submit button is disabled unless all fields are valid
- [x] validation errors aren't being highlighted to users
[...]

PR #113, opened 2018-05-15
Make buttons look more like GOV.UK buttons

pcraig3:
This pull request:

- Styles our buttons so that they look more like [GOV.UK buttons](https://govuk-elements.herokuapp.com/buttons/)
- Also exports the button styles so that we can use them on other elements that aren't buttons (ie, links for example [like on the landing page])
  - Finally, the link on our landing page is just a link now and not a link wrapping a button (this is great)
  - Adds `system-ui` as an explicit font declaration so that link buttons look identical to regular buttons
- Removes unicode arrows except on the first and last pages of our little service 

### screenshots

| before | after | mockup |
|--------|-------|--------|
| screen shot 2018-05-15 at 10 58 15 am |  screen shot 2018-05-15 at 10 58 54 am |   screen shot 2018-05-15 at 10 58 02 am   |

#### regular button
button

#### disabled button
button-2

<sup>note that we're going to enable this button soon (mentioned in #92), but this is just to make sure the disabled styles still work</sup>

### DIFF
--- a/web/src/CalendarPage.js
+++ b/web/src/CalendarPage.js
@@ -145,7 +145,7 @@ class CalendarPage extends Component {
               </CalReminder>
               <BottomContainer>
                 <Button disabled={submitting}>
-                  <Trans>Review →</Trans>
+                  <Trans>Review</Trans>
                 </Button>
 
                 <div>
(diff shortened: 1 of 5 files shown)


zakkain:
Looks great! Just one thing: we ended up going with no corner rounding on the buttons, so can you remove the border radius? We did that to match a lot of the other components, keeping everything sharp makes the visuals a lot more cohesive. 👍

pcraig3:
Okay, I updated the gifs but not the screenshoot

zakkain:
Killer, thanks! Merge it 💯

PR #114, opened 2018-05-16
Translations, moving files around, more focus

pcraig3:
This PR is sort of an un-reviewable grab bag

- Made the error message on the ErrorPage a bit shorter
- Added some super basic tests for the `FourOhFourPage` and the `ErrorPage`
- Added focus styles to links
- Fixed focus bug on buttons on Firefox
- Moved a bunch of `/src` files around
- Ran the translations extract + compile steps


pcraig3:
Only small changes in here, so 👍

PR #115, opened 2018-05-16
Catch errors more comprehensively

pcraig3:
Previously, our `Submission` component was catching GraphQL errors, like an email being sent with the wrong fields, but wasn't catching regular Errors, like a NetworkError.

Now it catches both.

### DIFF
--- a/web/src/components/Submission.js
+++ b/web/src/components/Submission.js
@@ -2,12 +2,8 @@ import React from 'react'
 import PropTypes from 'prop-types'
 import { Mutation } from 'react-apollo'
 
-const _logError = error => {
-  /* TODO log error.message */
-  return undefined
-}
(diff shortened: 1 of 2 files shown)


web/src/components/Submission.js
```diff
@@ -2,11 +2,17 @@ import React from 'react'
 import PropTypes from 'prop-types'
 import { Mutation } from 'react-apollo'
 
+const _logError = error => {
+  /* TODO log error.message */
+  return undefined
```
obrien-j: Not sure of the exact behavior in this case, but going on the assumption this is a 'pass', suggest we avoid this pattern, raise the error, take the crash, and more quickly implement the logging rather than relying on a TODO that might get missed down the road.

:2cents:

PR #117, opened 2018-05-18
Clean up content pages

pcraig3:
This pull request does a once-over for some of our pure-content pages. Mostly cleans up vertical spacing text sizes.
- Confirmation page is nicer
- 404 page is nicer
- Error page is nicer
- flag in footer is red
- "contact IRCC" information is standardized on different pages
  - the phone number is [a "tel:" link](https://css-tricks.com/the-current-state-of-telephone-links/) on smaller screens (for phones) but not on larger screens (because desktop computers typically don't have the capacity to make phone calls)
  - alternatively we could leave this as static text all the time if we don't think it's useful

@zakkain, it would be good if you would give the content a review. The confirmation page is in the mock-up so I put in the latest content, but for the error page and the 404 page I just took a best guess.

## confirmation page

[https://rescheduler.cds-snc.ca/confirmation](https://rescheduler.cds-snc.ca/confirmation)

| before | after |
|--------|-------|
|  image |   image   |

## error page

[https://rescheduler.cds-snc.ca/error](https://rescheduler.cds-snc.ca/error)
You can get to this by going directly to [the review page](https://rescheduler.cds-snc.ca/review), and trying to submit (you will get rejected because all of your data is empty).

| before | after |
|--------|-------|
|  image   |  image |

## 404 page

[https://rescheduler.cds-snc.ca/elephant](https://rescheduler.cds-snc.ca/elephant)

| before | after |
|--------|-------|
|   image  | image |


pcraig3:
Okay sweet 👍 

Yeah, updating the content is easy enough to do so it make sense to do this first.

PR #124, opened 2018-05-31
Add error tracking with Sentry

sleepycat:
This commit adds error tracking. It uses a new ErrorBoundary component from
gcui to catch errors occuring in the coponent tree and display an error state.
Sentry is used to log the error with an optional form for capturing user
feedback.

The ErrorPage has been changed slightly to be a little friendlier and more generic:

screenshot from 2018-05-31 11-56-12

Clicking the link brings up Sentry's user feedback form:

screenshot from 2018-05-31 11-56-22

### DIFF
--- a/web/src/components/Layout.js
+++ b/web/src/components/Layout.js
@@ -8,6 +8,8 @@ import PageHeader from './PageHeader'
 import AlphaBanner from './AlphaBanner'
 import FederalBanner from './FederalBanner'
 import Footer from './Footer'
+import { ErrorBoundary } from '@cdssnc/gcui'
+import ErrorPage from '../pages/ErrorPage'
 
 injectGlobal`
   html, body {
@@ -55,6 +57,14 @@ injectGlobal`
 
 const Layout = ({ children, contentClass = '' }) => (
   <div>
+    <ErrorBoundary
+      onError={(error, errorInfo) => {
+        window.Raven.captureException(error, {
+          extra: errorInfo,
+        })
+      }}
+      render={() => <ErrorPage />}
+    >
     <div role="banner">
       <AlphaBanner>
         <span>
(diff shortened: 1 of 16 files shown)


web/src/server.js
```diff
@@ -39,6 +39,8 @@ server
       })
       res.send(html)
     } catch (error) {
+      // eslint-disable-next-line no-console
+      console.log(error.message, error.stack)
```
pcraig3: Does this get passed through to a stack trace in Sentry?

web/src/components/Layout.js
```diff
@@ -55,6 +57,14 @@ injectGlobal`
 
 const Layout = ({ children, contentClass = '' }) => (
   <div>
+    <ErrorBoundary
+      onError={(error, errorInfo) => {
```
pcraig3: Are `error` and `errorInfo` something that we set in this code, or are they comeing from somewhere else?

web/src/server.js
sleepycat: No, this is just to help the debugging process. All that was happening in that catch was sending back error.message. That wasn't much to go on, so I thought at printing something to stdout would be good.

web/src/components/Layout.js
sleepycat: ErrorBoudary  components define a method `componentDidCatch`. That method gets called with `error` (a normal JS error object) and `errorInfo` (the component tree that produced the error). Those things are just passed on so they can be dealt with by the user of the component.

web/src/server.js
pcraig3: Okay, just curious. Seems okay to me.

web/src/components/Layout.js
pcraig3: Right, and in our case, we're passing the `errorInfo` to sentry so that it ends up in the Sentry logs, right?

PR #125, opened 2018-06-01
Mobile styling

nmakuch:


web/src/components/Summary.js
```diff
@@ -89,23 +80,24 @@ SelectedDayList.propTypes = {
   selectedDays: PropTypes.array,
 }
 
-const SummaryRow = ({ firstColumn, secondColumn, thirdColumn }) => (
+const SummaryRow = ({ firstColumnA, firstColumnB, secondColumn }) => (
```
pcraig3: Since we've moved away from columns, can we call them by the types of content they represent?

So maybe 
`firstColumnA` -> `SummaryHeader`
`firstColumnB` -> `SummaryBody`
`secondColumn` -> `SummaryLink`

or something similar?

pcraig3: Maybe even just `header`, `body`, `link`.

pcraig3:
About adding a test for our `ErrorList` component, here's what I think.

what you're doing already ([switching between emotion css blobs](https://github.com/cds-snc/ircc-rescheduler/blob/c1c721cd8c2fe201ea45888b8f98e82a2b0f0041/web/src/components/ErrorMessage.js#L35-L39)) is essentially the right idea, but it's not testable.

Emotion generates random-looking classnames when you add a `className={css`color: red`;}` declaration, like `.css-15isgng`. 
This means that we can't write a test that actually looks for different class names because we don't know what they will be beforehand.

To get around this with the `ErrorMessage` -- which is also supposed to be invisible until it has a message -- I added a class name as a string. Check it out here: https://github.com/cds-snc/ircc-rescheduler/blob/master/web/src/components/ErrorMessage.js#L59

That logic will add the `empty` class to an ErrorMessage with no `this.props.message` attribute.

That meant that I could write a test that
- [passes in no message and looks for the `.empty` class](https://github.com/cds-snc/ircc-rescheduler/blob/master/web/src/components/__tests__/ErrorMessage.test.js#L24)
- [passes in a message and checks that there is no `.empty` class](https://github.com/cds-snc/ircc-rescheduler/blob/master/web/src/components/__tests__/ErrorMessage.test.js#L36)

I think for the `ErrorList`, we should check a few things.

- Check that it renders without a message
  - should be a div with no contents, and with the `.empty` class
- Check that it renders with a message but without children
  - should be a div with a header and a paragraph
  - the message should be in the header
  - there should be an `.empty` class
- Check that it renders with a message and with children
  - should be a div with a header and a paragraph
  - the message should be in the header
  - there should be no `.empty` class
  - the children should be in the list

PR #138, opened 2018-06-07
Revert yarn.lock to fix hot reloading

pcraig3:
This is a bandaid more than anything.

When we run `yarn upgrade razzle`, our hot reloading is breaking. It might be to do with emotion, but I'm not sure.

Short term: I've upgraded all the other packages but reverted the razzle-specific ones. 
Long term: We upgrade razzle. Either it is fixed by itself (unlikely), we fix hot-reloading, or we lose hot-reloads.

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -31,7 +31,7 @@
     "lingui-react": "^1.4.1",
     "moment": "2.22.0",
     "raven-js": "^3.25.2",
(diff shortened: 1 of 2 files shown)


pcraig3:
Rolling back `eslint` introduces one low severity vulnerability.
Don't think this is a problem for us because `eslint` is something we only use in development. 

image

Issue #129, opened 2018-06-06
Calendar page function: date selection ordering


emanelfy:
## Date selection should change from being displayed in order of user selection to chronological order.
This will enable us to remove any unintended hierarchy associated with the date selection order, and present the dates back to the user chronologically for easier comprehension. 
<img width="640" alt="screen shot 2018-06-06 at 2 48 09 pm" src="https://user-images.githubusercontent.com/35348560/41059024-636dfa60-6999-11e8-8560-153a93330d09.png">

pcraig3:
This is in since #146 was merged.

PR #146, opened 2018-06-08
Order list of selected days chronologically

pcraig3:
Up to now, dates would show up in the order the user selected them.

ie, if they selected "June 8, June 5, June 1", they would be displayed in that order.

This could lead to the assumption that their preferred ordering was important when making their selection.

Since we're not letting them order dates at all, we're enforcing a chronological order which is how IRCC will get them.

now, if they select
-  "June 8, June 5, June 1"

they will see
- "June 1, June 5, June 8" 

as the dates they have selected.

Test it out here (might take a while to load but just grit your teeth): https://ircc-rescheduler-pyizkhsyuf.now.sh/calendar

## gif

dates

Merging this PR resolves issue #129

### DIFF
--- a/web/src/components/Calendar.js
+++ b/web/src/components/Calendar.js
@@ -373,14 +373,22 @@ const selectedDaysError = css`
   }
 `
 
+const sortSelectedDays = selectedDays => {
+  // create a new array because .sort() modifies our original array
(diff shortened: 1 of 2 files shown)


pcraig3:
gonna release this 👍

PR #147, opened 2018-06-11
Footer restyling

nmakuch:
**This PR consists of:**

| Before | After |
|--------|-------|
|    41063620-26eda6cc-69a7-11e8-9e2b-f991575f1376    |   screen shot 2018-06-11 at 11 59 19  |

**For Mobile:**
- Reduced the size of the Canada Wordmark by 50%
- Removed row-reverse as the flex-direction so that the footer is ordered Wordmark -> Links
- Font-size reduction on smaller screens for the links
- 'Terms and Conditions' becomes 'Terms' on smaller screens

**For Desktop:**
- Removed row-reverse as the flex-direction so that the footer is ordered Wordmark -> Links

### DIFF
--- a/web/src/components/Footer.js
+++ b/web/src/components/Footer.js
@@ -19,7 +19,6 @@ const footer = css`
   padding: ${theme.spacing.xl} ${theme.spacing.xxxl};
   display: flex;
   justify-content: space-between;
-  flex-direction: row-reverse;
   position: relative;
   font-size: ${theme.font.md};
 
@@ -40,12 +39,11 @@ const footer = css`
 
   ${mediaQuery.md(css`
     align-items: center;
-    flex-direction: row;
     padding: ${theme.spacing.xl} ${theme.spacing.xxxl};
   `)};
 
   ${mediaQuery.sm(css`
-    padding: ${theme.spacing.xl} ${theme.spacing.xl};
+    padding: ${theme.spacing.lg} ${theme.spacing.xl};
   `)};
 `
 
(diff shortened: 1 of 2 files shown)


web/src/components/Footer.js
```diff
@@ -19,23 +19,31 @@ const footer = css`
   padding: ${theme.spacing.xl} ${theme.spacing.xxxl};
   display: flex;
```
pcraig3: Since we're not using any flexbox rules anymore, we don't really need to keep this as `display: flex;`.

_But_, on the other hand, it doesn't hurt. So it's also fine to just leave it.

web/src/components/Footer.js
```diff
@@ -59,16 +71,22 @@ const bottomLinks = css`
...
   `)};
 `
+const terms = css`
+  span::after {
+    content: "Terms and Conditions";
+    ${mediaQuery.md(css`
+      content: "Terms";
+    `)};
+  }
+`
```
pcraig3: Generally, we want to avoid putting real-live content that people need in as CSS. [This stackoverflow answer](https://stackoverflow.com/a/47452364/9728185) is pretty succinct about it.

> CSS-generated contents may or may not be spoken depending on the browser and screen reader used, so you don't know for sure whether it will be spoken or not (don't use this to give important information).

CSS `content:` is great for putting in icons like arrow symbols or "x"es or whatever, but real content should always go into the HTML.

[We have some `visuallyhidden` CSS styles](https://github.com/cds-snc/ircc-rescheduler/blob/master/web/src/styles.js#L113-L126) in `styles.js`.

You could do something like this:

```js
import { visuallyhidden } from './styles'

const visuallyHiddenOnMobile = css`
  //mediaQuery small {
    ${visuallyhidden};
  }

<a>Terms<span className={visuallyHiddenOnMobile}> and Conditions</span></a>
```

☝️that should keep the words on the page all the time (better for accessibility), but hide the words we don't need on smaller screens.

PR #171, opened 2018-06-20
Confirmation email

timarney:
Staff a confirmation email:
- Removed hard coded template literal version of email
- Added ability to render variables into html / plain templates + send via params to mail service
- Added email param to GraphQL Mutation 
- Updated existing test for Mail (GraphQL)+ added tests to ensure markup is rendered
- Added inline-css to render styling for the email templates inline

Note: Will need to add French Versions to the existing templates


api/src/email.js
```diff
@@ -0,0 +1,117 @@
...
+function humanReadable(dates) {
+  return dates.map(date =>
+    new Date(date).toLocaleDateString('en-CA', {
+      weekday: 'long',
+      year: 'numeric',
+      month: 'long',
+      day: 'numeric',
+    }),
+  )
+}
```
pcraig3: Once we translate this email, we're gonna need a `lang` param in here or something.
(but for now it's cool!)

api/src/schema.js
```diff
@@ -60,47 +61,17 @@ const createSchema = t => {
...
-                  `,
-                },
-              },
-              Subject: {
-                Charset: 'UTF-8',
-                Data: 'IRCC Citizenship Rescheduling Tool',
-              },
-            },
-            Source: sendingAddress,
-            ReplyToAddresses: [sendingAddress],
```
pcraig3: 🎉  this is great

api/src/email_templates/staff-rich.html
```diff
@@ -0,0 +1,149 @@
...
+        margin: 0 auto;
+        display: inline-block;
+        margin: 0 auto;
+        background-color: #d8d8d8;
+        padding: 10px 20px;
+        color: #000;
+        text-decoration: none;
+        border: 1px solid #000;
+        font-weight: bold;
+    }
```
pcraig3: A bunch of this CSS isn't being applied anymore so let's get rid of it.

api/src/email_templates/staff-rich.html
```diff
@@ -0,0 +1,149 @@
...
+                <div>
+                    ${datesHtml}
+                </div>
+            </div>
+
+          
+
+
+
+
```
pcraig3: It might be nice to clear up some of this whitespace also. 
I know it's not a problem for the rendered email itself but it might be a bit cleaner just.

Up to you!

api/src/types/RescheduleForm.js
```diff
@@ -15,6 +15,10 @@ module.exports.default = t => {
         description: t('types.rescheduleForm.fields.fullName'),
         type: new GraphQLNonNull(GraphQLString),
       },
+      email: {
+        description: t('types.rescheduleForm.fields.email'),
+        type: new GraphQLNonNull(GraphQLString),
```
pcraig3: Do we want any stronger email validation? There are some vendor libraries (like [this one](https://www.npmjs.com/package/graphql-custom-types)) we could pull in to handle email validation for us if we wanted to reject bad emails at the API level.

What do you think?

api/index.js
```diff
@@ -12,13 +12,16 @@ if (!process.env.IRCC_RECEIVING_ADDRESS)
...
+if (!process.env.SITE_URL)
+  throw new Error('SITE_URL was not found in the environment')
 
 AWS.config.update({ region: process.env.AWS_REGION })
 
 const server = Server({
   mailer: new AWS.SES({ apiVersion: '2010-12-01' }),
   receivingAddress: process.env.IRCC_RECEIVING_ADDRESS,
   sendingAddress: process.env.SENDING_ADDRESS,
+  siteUrl: process.env.SITE_URL,
```
pcraig3: This is required for `inline-css` eh? We could get by at present with [a hardcoded empty string url](https://www.npmjs.com/package/inline-css#optionsurl) since we don't have any links (I think?) but do you know if we're needing them later? (I haven't looked at the email user mockup).

api/src/types/RescheduleForm.js
obrien-j: Tested that custom email regex with my john.o'brien@blah.com address... It doesn't fail, so lgtm. Might want to have a fallback 'This doesn't appear to be a valid email address, Send anyway' just in case the regex fails, then log/monitor regex failures.

api/index.js
timarney: Ya required for that lib.  I don't think there is a mock-up for the user email.  My understanding is we're just not showing some of the fields.  

As far a I can see we could do an empty string.  Thinking leave it for now until I see about getting that image embedded.

api/src/types/RescheduleForm.js
timarney: @obrien-j can you send me that regex I'll add on the front end under a different ticket that I'm working on.

sleepycat: Plus one to @pcraig3's suggestion. Using graphql-custom-types is a good idea. Making input types as specific as possible in all cases keeps the schema as a powerful, easily auditable security boundary. Apollo Engine should catch and record errors. If there are opportunities to tighten up any of the other types, jump on them.

timarney:
- Squashed a few of those commits together 
- Cleaned up the css that wasn't needed anymore with the layout updates
- Added the custom type for email check

PR #176, opened 2018-06-21
applicant email

timarney:
- Added applicant templates
- 2 separate e-mails are now sent ... 1 for staff and 1 for applicant

Note: In this version only the response from the staff message is sent back.

Will be swapping out the current mailer for Nodemailer in a separate PR at which point I'll chnage the response type to include responses from both emails that get sent.


api/src/email_templates/applicant-rich.html
```diff
@@ -0,0 +1,52 @@
...
+
+    .label {
+        font-weight: bold;
+        width: 200px;
+        vertical-align: top;
+    }
+
+    .value {
+        max-width: 400px;
+    }
```
pcraig3: We can remove this CSS rule here, it looks like.

pcraig3:
Also, this is a small point but we changed our page headers on [the review page](https://rescheduler.cds-snc.ca/review) a while back so that:
- all headers are sentence case, not title case
  - (ie, `This is sentence case`, not `This Is Title Case`)
- no more colons `:`

Could you update the staff email template to be the same as well? 
I know it's technically not part of this PR, but it's just a few quick edits to the [labels](https://github.com/cds-snc/ircc-rescheduler/blob/master/api/src/email_templates/staff-rich.html#L42).

api/src/email_templates/applicant-rich.html
timarney: Yes tested the e-mails sending locally - the applicant e-mail was the one I showed you on my phone.

timarney:
I can make these formatting updates under another PR tomorrow when I swap out the mailer.  

I'll be editing the templates under that "ticket" as I'll be adding the embedded Canada Flag logo

PR #181, opened 2018-06-22
cancel page

timarney:
Adds cancel page
screen shot 2018-06-25 at 9 33 22 am

Chevrons
screen shot 2018-06-25 at 3 01 46 pm

### DIFF
--- a/web/src/components/Layout.js
+++ b/web/src/components/Layout.js
@@ -78,6 +78,7 @@ class Layout extends React.Component {
       <div>
         <ErrorBoundary
           onError={(error, errorInfo) => {
+            if (!window || !window.Raven) return
             window.Raven.captureException(error, {
               extra: errorInfo,
             })
--- a/web/src/components/Layout.js
+++ b/web/src/components/Layout.js
@@ -9,7 +9,7 @@ import PageHeader from './PageHeader'
 import FederalBanner from './FederalBanner'
 import Footer from './Footer'
 import { ErrorBoundary } from '@cdssnc/gcui'
-import ErrorPage from '../pages/ErrorPage'
+import { ErrorPageContent } from '../pages/ErrorPage'
 import { initGA, logPageView } from '../utils/analytics'
 
 injectGlobal`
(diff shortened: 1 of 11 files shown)


web/src/pages/ErrorPage.js
```diff
@@ -58,7 +50,7 @@ class ErrorPage extends React.Component {
         <NavLink to="/">
           ← <Trans>Home</Trans>
         </NavLink>
-      </Layout>
+      </React.Fragment>
```
pcraig3: So, since we have an `/error` url path, this means that if someone were to visit `/error` by manually typing in the URL, say, they get an unstyled page.

The way I see it, we could either:

1. have an `<ErrorPage>` which renders `<Layout><Error /></Layout>`, and keep the body of our error page in a new component

2. remove the `/error` page since we don't want people to visit it.

3. Keep it as-is because it's a weird edge case.

What do you think?

I would lean towards (1), but I don't want to drag out this story for too long if you think it's pointless.

web/src/components/Layout.js
```diff
@@ -78,6 +78,7 @@ class Layout extends React.Component {
       <div>
         <ErrorBoundary
           onError={(error, errorInfo) => {
+            if (!window || !window.Raven) return
```
pcraig3: Good call here. 👍

web/src/pages/RegistrationPage.js
```diff
@@ -137,6 +137,16 @@ const validate = values => {
...
+
+  if (!validateEmail(values.email)) {
+    errors.email = (
+      <Trans>
+        You need to provide an email address so we can send you a confirmation
+        message.
+      </Trans>
+    )
+  }
+
```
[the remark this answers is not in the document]
timarney: ya was from a rebase

web/src/pages/ErrorPage.js
timarney: I think I'll set this up so it exports 2 components ErrorPage (default export) + ErrorContent

ErrorPage will be wrapped with the Layout and use ErrorContent as a child.

ErrorContent will be exported separately so it can be used by Layout for the React Error boundary (don't think that will ever git hit given the error seems to get picked up by Node)

PR #187, opened 2018-06-25
Translate alt text and links

Amirroracle:
-Created a new method called translateText, was having issues with using the conventional way of getting the i18n text via ```i18n._`text` ``` with regards to hyperlinks. Now uses i18n._('string') which resolves this. 

Added i18n to all the alt-text and links I could find. Let me know if I missed any, may need to add the translations in a separate PR though.


pcraig3:
Closing this in favour of #221 as it's pretty out of date.
However, the idea was a good one so I've reused exactly the same logic.

image

PR #190, opened 2018-06-25
Replace Apollo Link State for all of our user-submitted values

pcraig3:
We want to support a no-JS flow through our app -- that is, we want our app to work for clients who aren't running or can't run JavaScript.

Among other things, this means we need to:
- replace our caching layer because Apollo Link State only saves values to the client (whereas we need to be able to decide what HTML to send back from the server)
- handle the default form behaviour of submitting a request to the server with a bunch of query parameters

**This PR** 

- replaces Apollo Link state for our user-submitted values (the registration form and the calendar page).
  - It uses [React context](https://reactjs.org/docs/context.html) as a global state that can be passed around to nested components
  - It saves the context as a browser cookie and re-reads it later to support persistence between hard refreshes or a no-JS flow through the app.
- removes most of the Apollo Link State config from `utils/createApolloClient.js`
- removes the `LinkStateDisplay` component that was only useful for testing local cached values

**This PR does not**
- Redirect after successfully submitting your values on the registration page without JS. 
  - If you submit bad data, you will get validation messages
  - If you submit good data, you will just be returned to the page
- Add a straightforward way for users to clear out their session (ie, this should work with the 'cancel request' button)
- Make any consistent decisions about how long to save cookies for
- Remove all uses of Apollo Link State: we still have the language switcher to convert
- Make the calendar page work without JS. That is another story.
- Add any new tests. This totally sucks: this is testable code, but we're pressed for time.

**To test this PR**

Install new dependencies, and then build and run the app.
You should be able to do all of the things you can currently do: make it back and forth through the app with your values being saved between pages.

New things you can do:
- on the `/register` page, you can hit "continue" without JS running on the page
  - any values that pass validation will be saved
  - any values that fail validation will cause error messages to show up when the page reloads
- from any point in the app, you can do a hard refresh and not lose your previously saved values

**To learn more**

Reading the diff for this PR is terrible because there's a lot of highlighted line diffs that came about from un-indenting the nesting in various places. The actual code changes are hard to find amongst the whitespace changes.

A simplified version of all this code can be found at my [after-no-js](https://github.com/pcraig3/after-no-js) repo where I trialled all this stuff.
- read [REQUEST_LIFECYCLE.md](https://github.com/pcraig3/after-no-js/blob/master/REQUEST_LIFECYCLE.md) for an overview
- build the project and click around 
  - if you uncomment the console statements, you will see when it sets and reads cookies
- ask me about it


timarney:
This is great - overall I think it would be good to outline for clarity / debugging  what takes priority on a load / refresh i.e. URL params vs Cookie (if both exist) also does this change in a JS / no JS env.

- French Toggle -  JS doesn't seem to handle the toggle
- Not sure steps to replicate but travel radio disappears (something to do with flipping JS off / on?)
- If the cookie exist should that also reload the data even with JS off?  if yes I seem to be having an issue 


screen shot 2018-06-25 at 3 50 50 pm

timarney:
## Testing

**Full Happy Path**
- JS -> Full run through (good values)
- No JS -> Full run through (good values)

**Partial Happy Path** 
- JS -> Full run through (test validation of inputs)
- No JS -> Full run through (test validation of inputs)

**Delayed Happy Path**
- JS ->Run through partial - i.e. user had to take care of something ... walks away (comes back)
- No JS -> Run through partial - i.e. user had to take care of something ... walks away (comes back)

pcraig3:
Forgot to configure cookie parsing in the server, so the cookies weren't being found on the first load. Updated in https://github.com/cds-snc/ircc-rescheduler/pull/190/commits/e84dd9314fba5857ae1d35341b09593f3434623a

As to your points:
> French Toggle - JS doesn't seem to handle the toggle

Yeah, the language toggle isn't hooked up to this yet.

> Not sure steps to replicate but travel radio disappears (something to do with flipping JS off / on?)

This is has been a persistent issue to do with using our radio buttons in [`react-final-form`](https://github.com/final-form/react-final-form). Have to dig into it a bit to see what's up.


> If the cookie exist should that also reload the data even with JS off? if yes I seem to be having an issue

Yes, but it was the server config issue. Now we should be a-okay.

pcraig3:
tenor

PR #200, opened 2018-06-26
Add updated translations and compile all content again

pcraig3:
This is so that we have all the updated content in our `messages.json` file for tomorrow.


pcraig3:
image

PR #203, opened 2018-06-27
Consolidate cookie validation logic + add `language` to store

pcraig3:
There's a lot going on here, but, as before, it's been prototyped out at https://github.com/pcraig3/after-no-js. This PR doesn't update our language switcher to actually sent requests to change the language, but you can see this working at [https://nojs.now.sh/](https://nojs.now.sh/).

## Summary of changes:

1. Consolidate cookie validation logic

Both server-side and client-side cookies are put through the same validation logic now, in `WithProvider.validateCookie()`.
Before, the No-JS validation was happening in [`setSSRCookie`](https://github.com/cds-snc/ircc-rescheduler/blob/master/web/src/cookies.js#L59), while the JS validation was happening in the pages themselves. They were similar but now they are the same.
Also added a bunch of tests for it since it's very easy to throw things at it and see what comes back.

2. Add `language` as a globally-settable field

Previously, we were able to get saved values back on any page, but we could only set them on specific pages.
Since we can send a server-side language change request (?language=fr) from any page in the flow, we need to be able to set a field globally.
Since the pages themselves have a `static get fields()` function that lists permissible fields for that page, I added a `static get fields()` function to `WithProvider` so that all query strings coming through can be checked for whether they contain the `language` query parameter.
The validation logic for this is really gross looking so I added a bunch of tests for that as well.

3. cookies.js is much simpler now

Took all checks out of these methods now that everything is in `WithProvider.validateCookie()`

4. Added (optional) secret env var: RAZZLE_COOKIE_SECRET

If we set a 32-character random string in our .env.production file, we will have a real secret and not one that's public in our github.

5. Shorten cookie expiry time in dev mode

Cookies are only set to a week long in production.
In dev mode they're only 10 minutes.

### DIFF
--- a/web/src/__tests__/WithProvider.test.js
+++ b/web/src/__tests__/WithProvider.test.js
@@ -1,0 +1,212 @@
+import withProvider from '../withProvider'
+
+class FakeComponentEmpty {
+  constructor() {}
+}
+
+describe('WithProvider', () => {
+  describe('.globalFields and .validate()', () => {
+    const WithProvider = withProvider(FakeComponentEmpty)
+
+    it('has only "language" as a global field', () => {
+      expect(WithProvider.globalFields).toEqual(['language'])
+    })
+
+    it('returns no errors for "en" or "fr" in its validation method', () => {
+      let errors
+      errors = WithProvider.validate({ language: 'en' })
+      expect(errors).toBe(false)
+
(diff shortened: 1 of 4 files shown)


pcraig3:
tenor

🚢 🚢 🚢

PR #204, opened 2018-06-27
Cancel clear

timarney:
Adds clear functionality to clear user data (cookies) when they press the cancel button. 

Prior functionality took the user to the cancel page but didn't clear the user data.

### DIFF
--- a/web/src/server.js
+++ b/web/src/server.js
@@ -26,6 +26,10 @@ server
     res.clearCookie('store')
     return res.send('no more cookies 🍪')
   })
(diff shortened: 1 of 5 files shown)


web/src/server.js
```diff
@@ -24,7 +24,7 @@ server
   .get('/clear', (req, res) => {
     /* TODO: this needs improvement */
```
pcraig3: We can get rid of this now, but otherwise let's ship it 🚢

PR #218, opened 2018-06-29
Make a GraphQL post from the server II

pcraig3:
A couple changes in here:

- created a new object from `res.body` so that we're not mutating it
- added paperFileNumber to test data
- removed 'test' mutation since we no longer need it
- redirect to `/confirmation` if no error is thrown
- added an absolute URL in production since that seems to be an issue


pcraig3:
image

PR #219, opened 2018-06-30
Merge 1st: Reveal radios + checkboxes

pcraig3:
This PR does a couple things:

- removes some unused CSS from the checkboxes + radios styling
- cleans up a linting error we were getting by not passing an array to `oneOfType`
- adds some dummy input elements to our registration page and No-JS calendar page
  - not my proudest moment (more info below)

***

## screenshots

| before | after |
|--------|-------|
|  radio hidden on the registration page   |  radio unhidden!    |
|     screen shot 2018-06-29 at 9 18 35 pm   |    screen shot 2018-06-29 at 9 18 37 pm   |
|  small checkboxes + first one is hidden  |  larger checkboxes + first one unhidden!    |
|   screen shot 2018-06-29 at 9 19 54 pm   |  screen shot 2018-06-29 at 9 20 00 pm     |

***

### Add dummy radio button / checkbox to our pages 
This is nuts.

Pseudo element styles are not being applied correctly to the first
radio button and checkbox on the page, when in No-JS mode.

I don't know why this is happening really, but I don't want to
debug it at this point.

Adding a dummy radio to the registration page and a dummy checkbox
to the No-JS calendar page and then wrapping them in a hidden div.
- 'display: none' hides elements from visual interfaces and AT
- they are outside of our forms so they can't be submitted

### DIFF
--- a/web/src/components/CalendarNoJS.js
+++ b/web/src/components/CalendarNoJS.js
@@ -62,18 +62,16 @@ const Calendar = ({ startDate, endDate, dates }) => {
       const checked = dates.includes(val)
 
       const el = (
-        <React.Fragment key={val}>
-          <li>
-            <Checkbox
-              name="calendar"
-              id={`calendar-${idMonth}-${index}`}
-              value={val}
-              label={label}
-              onChange={() => {}}
-              checked={checked}
-            />
-          </li>
-        </React.Fragment>
+        <li key={val}>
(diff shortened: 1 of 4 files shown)


pcraig3:
Yeah, and then open up an issue on the repo if possible.
I agree that that would be the not-insane way to solve this, and we should make some time for it.

Issue #161, opened 2018-06-18
a11y issues 


JuliannaR:
Here is a quick list of issues related to the reschedule project. Please note that this is not an exhaustive list. (time spent: 20 minutes/SME + 10 minutes/user testing) 

Meth: Tested with WAVE, Nu Validator, Colour Contrast Analyzer, keyboard and user tested for adaptive technology (screen reader and voice rec).

1. Contrast Minimum:  Buttons fail (Start, Continue, Review) - Can be testing with colour contrast analyzer.

![cdsreschcolour](https://user-images.githubusercontent.com/35537308/41546727-69911ee8-72ec-11e8-97c6-47846830dfc2.JPG)

2. Info and Relationships: Headings
        a. https://rescheduler.cds-snc.ca/ Request a new Canadian Citizenship appointment should be an H1. 
        b. please do not use hidden headings. Re: First verify your identity and tell us why you need a new  
            appointment. 

3. Language of Page: Language declaration is not set on pages. 

4. Error Identification: If you select only 1 or 2 dates in the calendar there error description is not sufficient to identify the issue. I would suggest indicating that 3 date selections are required and indicate that it is required.

     "Sorry, there was a problem with the information you submitted.
      Please take another look at the:

      Calendar"

5. Info and Relationships/Labels and Instructions: Remove date button is not associated with which date is being removed. This causes a burden and makes it difficult for a user to understand which dates they are removing or adding. 

6. Link Purpose: https://rescheduler.cds-snc.ca/review and https://rescheduler.cds-snc.ca/calendar 
   Adjacent links go to the same URL. change, change, change, and review, review, review

![cdsreschadjacentlinks](https://user-images.githubusercontent.com/35537308/41546687-50d5b9e0-72ec-11e8-9890-2df2a264270c.JPG)

   When adjacent links go to the same location (such as a linked 
   product image and an adjacent linked product name that go to the same product page) this results in 
   additional navigation and repetition for keyboard and screen reader users.

Usability Findings:

Chrome & JAWS 2018
Trying to correct a date seems to crash the calendar. You need to start over completely when clicking Remove date.

IE and JAWS 2018 
July seemed fine until I clicked the remove button on one of the dates. It erased ALL of them.

## Issues

- [x] Contrast Minimum: Buttons fail
- [ ] Info and Relationships: Headings
- [x] Language of Page
- [x] Error Identification
- [x] Info and Relationships/Labels and Instructions
- [x] Link Purpose

[...]

PR #220, opened 2018-07-02
Merge 3rd: Server-side language-switching

pcraig3:
# User needs

- As a user, I can click a link and get a server-rendered french site, with or without JS.
- As a user, I can set a language preference, refresh the page, and still have my preferred language.

# High-level technical details

This PR switches over to our React context + browser cookies strategy as the source of truth for our rendered language. 

It also moves `<I18nProvider>` into `withProvider`, and adds `withProvider` to all pages by default so that all pages or their descendent components have access to the Context if they need it.

Since language was the last thing we needed Apollo Link State for, this PR also removes ApolloProvider as much as possible from our application. 
It's still needed to `POST` to the email API, but other than that it's been removed.

This PR also sets the "lang" attribute in the `<head>` based on the current language. This was one of the issues noted in #161, but we weren't able to solve it until now because we were always sending back content in english.

_There's much more detail in the various commit messages, so read those for a more granular overview._


web/src/pages/__tests__/CancelPage.test.js
```diff
@@ -4,7 +4,7 @@ import React from 'react'
 import { MemoryRouter } from 'react-router'
 
 describe('<CancelPage />', () => {
-  it('can be instantiated', () => {
+  it.skip('can be instantiated', () => {
```
timarney: We should remove this skip

web/src/pages/__tests__/ErrorPage.test.js
```diff
@@ -13,7 +13,7 @@ describe('<ErrorPageContent />', () => {
 })
 
 describe('<ErrorPage />', () => {
-  it('can be instantiated', () => {
+  it.skip('can be instantiated', () => {
```
timarney: We should remove this skip

web/src/pages/__tests__/CancelPage.test.js
pcraig3: The tests don't run, i don't think, until enzyme updates itself.

https://github.com/cds-snc/ircc-rescheduler/pull/220/commits/3ff1fe7733fa26fdf5542adcaa94a32dbd52cd89

pcraig3: But we should fix this ASAP.

PR #221, opened 2018-07-02
Merge 4th: Translate alt text, links, and page titles 

pcraig3:
**Note: this PR follows up on #220, which should be merged first.**

***

# User needs

- As a french-speaking user, I can click links to french content offsite
- As a french-speaking user with a screenreader, the alt-text should update when I select a new language preference
- As a french-speaking user with a screenreader, the page title should update when I select a new language preference 

# Technical details

_This PR is only from commit 2e53f37d1439e60fb2341eb9efc129f9a897db7c onwards._

Following up on #220, we can now get back server-rendered pages in French as well as English. However, this wasn't working in places where we needed to use static text: such as alt text, links, or page titles.

Now it _does_ work. I've done pretty much the same as what @Amirroracle did in his PR #187, but I'm closing that PR since it's pretty out of date at this point.

An unfortunate side-effect of wrapping our components with `withI18n` is that we have to use `mount` instead of `shallow` for our tests. Anyway, as long as we know what we're dealing with.


web/src/components/__tests__/Footer.test.js
```diff
@@ -1,21 +1,41 @@
 import React from 'react'
-import { shallow, render } from 'enzyme'
+import { mount } from 'enzyme'
 import { FooterBase as Footer } from '../Footer'
 import { getStore } from './LanguageSwitcher.test.js'
 
 describe('<Footer />', () => {
   it('renders footer', () => {
-    const footer = shallow(<Footer context={getStore('en')} />)
+    const footer = mount(<Footer context={getStore('en')} />)
```
Amirroracle: Tested this out locally, should still work with render, not a big deal since you only shave off about 10 milliseconds.

web/src/components/__tests__/Footer.test.js
```diff
@@ -1,21 +1,41 @@
...
     expect(footer.find('footer').length).toBe(1)
     expect(footer.find('hr').length).toBe(0)
   })
 
   it('renders footer with topBar', () => {
-    // have to use 'render' instead of 'shallow' to render nested components
-    const footer = render(
+    // have to use 'mount' instead of 'shallow' to render nested components
+    const footer = mount(
```
Amirroracle: Same with here, render can be used here. Up to you if you want to keep it that way.

web/src/components/__tests__/Footer.test.js
```diff
@@ -1,21 +1,41 @@
...
       <Footer topBarBackground="black" context={getStore('en')} />,
     )
     expect(footer.find('footer').length).toBe(1)
     expect(footer.find('hr').length).toBe(1)
   })
+
+  it('renders "and Conditions" in English', () => {
+    const footer = mount(<Footer context={getStore('en')} />)
```
Amirroracle: Though mount here is needed so 👍

web/src/components/__tests__/Reminder.test.js
```diff
@@ -1,10 +1,10 @@
 import React from 'react'
-import { shallow } from 'enzyme'
+import { mount } from 'enzyme'
 import Reminder from '../Reminder'
 
 describe('<Reminder />', () => {
   it('renders reminder message', () => {
-    const reminder = shallow(<Reminder>Message</Reminder>)
+    const reminder = mount(<Reminder>Message</Reminder>)
```
Amirroracle: Same deal here, render() should work.

PR #226, opened 2018-07-03
Quick wins 🏆

### LINKED ISSUE #128: French translations omnibus
Our application must be translatable into both English and French.
This is currently working imperfectly, but there are several issues preventing us from offering the same experience in French that we do in English. 

Let's look at them.

# What we have now

Currently, we can link to [our service](https://rescheduler.cds-snc.ca/) and click the `Français` link at the top and the page content will instantly switch to French. 
This is an in-page change and it will switch back to English if the page is reloaded. 
It also does not work without JavaScript.

# What we want

## 1. Linking directly to French language-pages (with JavaScript)

This means having some capacity to link to something like [https://rescheduler.cds-snc.ca?lang=fr](https://rescheduler.cds-snc.ca?lang=fr) and have the page switch immediately to french. This is roughly the equivalent of loading a page and clicking the `Français` link immediately. 
This would mean:
- being able to send links to french-language pages 
- reloading a page would not cause users to lose their set language

However, this would only work on JS-enabled devices.

**This is a medium-sized story.**

## 2. French-language pages without JavaScript

This means our pages would come back from the server in French. This would allow us to fall back to doing a full page reload in French if non-JS users clicked the `Français` link. 

This would also solve our `lang=“en”` problem, which tells screenreaders to expect which language the content will be on the first load.

**This is a large-sized story, and needs some exploration.**

## 3. Translating links and alt text

Our links don't change to French ones, as pointed out by @lchski in [this (superseded) issue](https://github.com/cds-snc/ircc-rescheduler/issues/121). 
Our alt-text doesn't change either.

Changing HTML attributes is different from changing the page content, so we need to investigate this. [There's some documentation around this](https://lingui.github.io/js-lingui/guides/optimized-components.html?highlight=attributes#optimized-component-wrapped-in-withi18n), but I haven't read it very closely.

**This is (probably) a medium-sized story.**

## 4. We lose text formatting when we switch languages.

| before | after |
|--------|-------|
|  "paper file number" is bolded |  "numéro de fichier paper" is not bolded   |
|  screen shot 2018-06-04 at 9 42 01 pm  |  screen shot 2018-06-04 at 9 41 43 pm    |

Whenever our markup looks like this 👇, we lose whatever was in the `<strong>` tags.
```html
<p>
  <Trans>
    Citizenship Tests are scheduled on <strong>Tuesdays</strong> and <strong>Fridays</strong>.
  </Trans>
</p>
```
The fix for this is fairly trivial, but it'll take some time.
We should probably do is closer to the end when the content is more finalized.

**This is a small-sized story.**

## 5. The calendar is not translated

It looks like one thing, but it breaks down into a few different small tasks.
Related to the Calendar component itself, [there's documentation here](http://react-day-picker.js.org/). Specifically, we probably want to check out [the locale stuff](http://react-day-picker.js.org/api/LocaleUtils). 

Calendar settings
- Month name is not in French
- Weekday names are not in French
- The `aria-label` attribute for individual days is formatted in english

Date formatting
- Once days are selected using the calendar, the date that shows up in the box at the side should be formatted in French

Separately, each of these bullet points is probably a small story, but together they make up a medium-y type story.

**Taken together, this is a medium-sized story.**

### LINKED ISSUE #224: Bolding Sweep
### The following outlines the only things that should be bolded on each page:

--

### **Header**
- Request a new citizenship appointment

- Beta feedback text-weight should be 400 across the whole service. 

--

### **Landing**
- 3 days
- Warning box text: Requesting a new appointment will cancel your current one. Do not attend your old appointment after you complete this request. It can take up to 9 weeks for us to reschedule you.

--

### **Form**
Remains as is.

--

### **Calendar**
- 3 days
- Your 3 selected days:
- Warning box text: Make sure you stay available on all of the days you select.

--

###  **Review,** Confirmation, Cancel, and Something Went Wrong
Remain as is.

--
_Team members directly affected:_ @nmakuch @anikbrazeau @Amirroracle

pcraig3:
These turned into not very quick wins actually.

In this PR:
- Fixed a bug where the chevrons were too close to their links
- Make all the visited links blue
- Unbolded things where we wanted to unbold them
  - resolves #224 
- Translates our links
  - resolves #128 
- Preserves bolded text to the french version of the site
  - also resolves #128 
- Fixes a layout bug we were seeing with the noJS calendar, where the "selected days" box would always be underneath it.

Gonna executive decision merge this PR and then deploy it since this is all needed for our go-live.


pcraig3:
deploy

PR #227, opened 2018-07-04
Dynamic Dates for JS cal + translations for dates 

timarney:
- Created utility functions for dates (start ,end etc...) `calendarDates.js`
- JS cal dates are no longer hardcoded (now using `calendarDates.js`)
- Date strings now use locale and are translated via <Time> or makeGMTDate from `Time.js`
- Added ability to pass options to the <Time> component  i.e.  { month: 'long', year: 'numeric' }
- Calendar test we're not updated am I wrong thinking those will need a hard-coded date to test against?

@pcraig3 

Tried to get the `let { context: { store: { calendar = {} } = {} } = {} } = this.props` line working but ended up resorting to the following for now.

```
 if (
      this.props &&
      this.props.context &&
      this.props.context.store &&
      this.props.context.store.language
    ) {
      locale = this.props.context.store.language
    }
```

## Translated Dates

screen shot 2018-07-04 at 9 01 09 am

## Dynamic confirmation date

screen shot 2018-07-04 at 12 23 37 pm

## Start End Dates
Adjusted start and end dates to match the following:
Today + 6 weeks = first available dates
8 weeks presented as available in total

## JS Cal

screen shot 2018-07-04 at 12 35 21 pm

## Landing Page

landingpage


web/src/utils/calendarDates.js
```diff
@@ -0,0 +1,99 @@
...
+const offsetStartWeeks = 6
+// Count 8 weeks from that point (ie, add 56 days)
+const offsetEndWeeks = 8
+
+const offsetRespondBy = 6 // weeks
+
+export const getStartDate = (today = new Date(), parseToDate = false) => {
+  const date = wedOrThurs(addWeeks(today, offsetStartWeeks))
+  return parseToDate ? parse(date) : dateToISODateString(date)
+}
```
pcraig3: If the `parseToDate` parameter is only being used by the calendar, could we just handle that logic there?

Ie, in `Calendar.js`, we could do:

```
// we need to parse this because the calendar doesn't like something for some reason
before: parse(getStartDate(new Date()))
```

Or do you think it's better in here?

web/src/utils/calendarDates.js
```diff
@@ -0,0 +1,99 @@
...
+
+  return parse(addDays(date, i))
+}
+
+export const getStartMonth = (today = new Date()) => {
+  const baseDate = parse(getStartDate(today))
+  return startOfMonth(wedOrThurs(baseDate))
+}
+
+export const getStartMonthName = (today = new Date(), locale = 'fr') => {
```
pcraig3: ooh, bold move setting the default to `fr`, eh? 

😁

web/src/utils/calendarDates.js
```diff
@@ -0,0 +1,99 @@
+import addWeeks from 'date-fns/add_weeks'
+import parse from 'date-fns/parse'
+import isWednesday from 'date-fns/is_wednesday'
+import isThursday from 'date-fns/is_thursday'
+import startOfMonth from 'date-fns/start_of_month'
+import addDays from 'date-fns/add_days'
+import subWeeks from 'date-fns/sub_weeks'
+import format from 'date-fns/format'
+import { makeGMTDate, dateToISODateString } from '../components/Time'
```
pcraig3: By the way, if it makes more sense to take any of these methods out of here or rewrite them with `date-fns` stuff, that's totally possible.

I was keeping all the date formatting stuff in `Time.js` before because it was the only place that was dealing with dates. 

Now that we have this file, we could move stuff over if we needed to.

pcraig3: Fine with either, just thought I would mention it.

web/src/pages/ConfirmationPage.js
```diff
@@ -17,6 +20,27 @@ const contentClass = css`
...
+    let locale = 'en'
+
+    if (
+      this.props &&
+      this.props.context &&
+      this.props.context.store &&
+      this.props.context.store.language
+    ) {
+      locale = this.props.context.store.language
+    }
```
pcraig3: also ☝️

web/src/utils/__tests__/calendarDates.test.js
```diff
@@ -24,6 +25,16 @@ describe('Utilities functions CalendarDates.js', () => {
 
   it('gets to month', () => {
     const today = new Date('September 05, 2018')
-    expect(toMonth(today, false)).toEqual('2018-11-28')
+    expect(toMonth(today, false)).toEqual('2018-12-12')
+  })
+
+  it('gets confirmation date', () => {
+    const selectedDays = ['2018-01-31', '2018-01-30', '2018-08-22']
+    expect(respondByDate(selectedDays, 'en')).toEqual('July 11, 2018')
```
pcraig3: Not related to this test, but do you think we should just sort the selected dates array?

pcraig3: Regardless, we should have a test for this.

web/src/utils/__tests__/calendarDates.test.js
```diff
@@ -24,6 +25,16 @@ describe('Utilities functions CalendarDates.js', () => {
...
+  })
+
+  it('returns null if selectedDays value(s) not passed', () => {
+    const selectedDays = ['']
```
pcraig3: I think technically we start with `selectedDays = []`, not `selectedDays = ['']`.

The difference being that `[].length === 0` vs `[''].length === 1`. 

Probably this test works the same either way, but just so we're consistent.

web/src/pages/LandingPage.js
```diff
@@ -95,6 +108,8 @@ class LandingPage extends React.Component {
             <strong>
               <Trans>3 days</Trans>
             </strong>{' '}
+            <Trans>in </Trans> {getStartMonthName(new Date(), locale)}{' '}
+            <Trans>or</Trans> {getEndMonthName(new Date(), locale)}{' '}
```
pcraig3: I think `to` is more appropriate here.

Sometimes we'll have 2 months so it'll say: 
```
Then you'll select 3 days in August or September
```

But when we have three months, it's a bit weird to say
```
Then you'll select 3 days in August or October
```

I've asked @emanelfy and she suggests
```
Then you'll select 3 days between August to October
```

pcraig3:
Here, I'm going to test this out locally also just for a once-over.

web/src/utils/calendarDates.js
```diff
@@ -0,0 +1,99 @@
...
+
+// Go 6 weeks from today (ie, add 28 days)
+const offsetStartWeeks = 6
```
pcraig3: Hey, after talking to @emanelfy, she says we were off by one week. 

If we make this `5` instead of `6`, we should be good.

web/src/utils/calendarDates.js
```diff
@@ -0,0 +1,99 @@
...
+// Count 8 weeks from that point (ie, add 56 days)
+const offsetEndWeeks = 8
+
+const offsetRespondBy = 6 // weeks
```
pcraig3: Hillary and @emanelfy want to know if we can make an offset time that's like 4 weeks + 2 days?

web/src/components/Time.js
```diff
@@ -33,15 +34,16 @@ const dateToISODateString = date => {
...
   <time dateTime={dateToISODateString(date)}>
-    {dateToHTMLString(date, locale)}
+    {dateToHTMLString(date, locale, options)}
   </time>
 )
 Time.propTypes = {
   date: PropTypes.oneOfType([PropTypes.string, PropTypes.instanceOf(Date)])
     .isRequired,
   locale: PropTypes.string,
+  options: PropTypes.object,
```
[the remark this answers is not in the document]
pcraig3: It would be good to add a test in the time file for the new format you're using also.
We don't have to test JS's locale date formatting extensively or anything, but since we're outputting two different formats, a sense-check there would be good.

pcraig3:
Last thing I noticed (and I don't think we should let this hold up the PR): when you're in no-JS mode, the dates aren't formatting.

image

I think it's because command-line node doesn't ship with locale information by default, and then since the browser JS isn't turned on, the dates never get translated properly.

I'll write a bug for this in trello we should merge this PR without worrying too much about it.

timarney:
Interesting with the command line node stuff.  Is that the same type of bug you guys hit with Amir's code.

I'll make the adjustments in the a.m thanks for reviewing.

web/src/utils/calendarDates.js
timarney: Ya is that for offsetRespondBy?

web/src/components/Calendar.js
```diff
@@ -527,14 +530,14 @@ class Calendar extends Component {
           months={dateInfo.months}
           weekdaysLong={dateInfo.weekdaysLong}
           weekdaysShort={dateInfo.weekdaysShort}
-          initialMonth={new Date(2018, 7)}
-          fromMonth={new Date(2018, 7)}
-          toMonth={new Date(2018, 8)}
+          initialMonth={startMonth}
```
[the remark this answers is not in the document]
timarney: Right forgot we'll need that for page reload.  Would be first user selected day or default to calculated initial month correct?

web/src/components/CalendarNoJS.js
```diff
@@ -113,28 +123,31 @@ Calendar.propTypes = {
   startDate: isValidDateString,
   endDate: isValidDateString,
   dates: PropTypes.oneOfType([PropTypes.array, PropTypes.object]),
+  locale: PropTypes.string,
```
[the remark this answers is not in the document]
timarney: Will fallback to en as a default but will make it required as that makes more sense.

web/src/components/__tests__/Calendar.test.js
```diff
@@ -40,7 +45,7 @@ describe('<CalendarAdapter />', () => {
     expect(wrapper.text()).toMatch(/August 2018/)
   })
 
-  it('renders September 2018', () => {
+  it.skip('renders September 2018', () => {
```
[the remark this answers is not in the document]
timarney: Happy to take a look just a bit confusing as I have worked with the JS cal yet.

web/src/utils/calendarDates.js
timarney: ya good call - removed from the util function as it just makes things more confusing

web/src/utils/calendarDates.js
timarney: lol - figured if something was wrong we would notice right away

web/src/pages/CalendarPage.js
```diff
@@ -176,6 +185,18 @@ class CalendarPage extends Component {
...
+
+    if (
+      this.props &&
+      this.props.context &&
+      this.props.context.store &&
+      this.props.context.store.language
+    ) {
+      locale = this.props.context.store.language
+    }
+
```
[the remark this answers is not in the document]
timarney: This will click for me soon :)

web/src/utils/calendarDates.js
```diff
@@ -0,0 +1,99 @@
...
+    addWeeks(new Date(getStartDate(today)), offsetEndWeeks),
+  )
+
+  return endDate
+}
+
+export const wedOrThurs = date => {
+  let i = 0
+  //find the current or next Wed or Thurs
+  for (i = 0; i <= 7; i++) {
```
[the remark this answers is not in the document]
timarney: swapped to use var just need the one.

web/src/utils/calendarDates.js
timarney: Thinking quite a bit of this will get refactored if we make the move to multiple locations.  I think it's fine pulling from time for now.

web/src/utils/__tests__/calendarDates.test.js
timarney: Good call.  Added.

web/src/components/Calendar.js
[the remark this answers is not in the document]
pcraig3: Right now it defaults to the calculated initial month. 

But if the user has selected all days in September, say, and then goes to the review page and comes back, the calendar is still on "August 2018", and you can't see any of the days you've selected.  It would be nice to render the initial month as September in that case.

There's [a Trello card for this](https://trello.com/c/VTRk3OJl/198-when-you-select-days-on-the-calendar-that-are-not-the-initial-month-and-then-you-return-to-the-calendar-page-it-should-load-your) (with a [gif](https://trello-attachments.s3.amazonaws.com/59e8a4289cfe1c1d0b7c02bb/5b325a1af070e77910004074/6d30f8af22d47df1464329abf2446c3d/calendar-1.gif)), and it seems like a small use case, so we don't have to solve it now.

web/src/components/__tests__/Calendar.test.js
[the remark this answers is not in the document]
pcraig3: Okay, sure.

There's kind of two main things that will have to be made dynamic.

1. I've hardcoded all the date strings. So it would be good to have a function that can handle "get first available enabled day", "get 2nd available enabled day" and then get the string format we expect. I think we only use the first 4 or so.

2. There're some tests where I'm passing the calendar a date so that on first render it already has a date selected. So if we could dynamically get the first, second (etc.) dates -- we have to keep them as Date objects -- and then pass them to the calendar, that should solve that issue.

It's kind of the same logic in both cases, we just want the output format to be different.

web/src/components/__tests__/Time.test.js
```diff
@@ -22,6 +22,24 @@ describe('<Time />', () => {
...
+  it('renders with formated options month and year', () => {
+    const time = shallow(
+      <Time
+        date={dateString}
+        locale={'en'}
+        options={{ month: 'long', year: 'numeric' }}
+      />,
+    )
+    expect(time.text()).toEqual('April 1870')
+  })
```
pcraig3: 👍

web/src/pages/LandingPage.js
```diff
@@ -63,6 +66,16 @@ const LongReminder = styled(Reminder)`
...
+    let locale = 'en'
+
+    if (
+      this.props &&
+      this.props.context &&
+      this.props.context.store &&
+      this.props.context.store.language
+    ) {
+      locale = this.props.context.store.language
+    }
```
pcraig3: Could also blow this up as a deconstruction thing, but we don't have to. 🤷‍♂️ 

Would look a lot like the others.

web/src/utils/__tests__/calendarDates.test.js
```diff
@@ -0,0 +1,45 @@
...
+
+  it('gets to month', () => {
+    const today = new Date('September 05, 2018')
+    expect(toMonth(today, false)).toEqual('2018-12-05')
+  })
+
+  it('gets confirmation date', () => {
+    const selectedDays = ['2018-01-31', '2018-01-30', '2018-08-22']
+    expect(respondByDate(selectedDays, 'en')).toEqual('July 27, 2018')
+  })
```
pcraig3: If you pass in `'fr'`, this will translate, right? (Our jest tests should be working with date-formatting locales)

You could add a quick test for that?

web/src/utils/calendarDates.js
pcraig3: 👍

web/src/utils/calendarDates.js
pcraig3: haha, good point.

Issue #191, opened 2018-06-25
Calendar page: date selection redesign


emanelfy:
**Overall Changes:**
- [ ] date buckets
- [ ] calendar grey
- [ ] on-state selection
- [ ] bolding
- [ ] mobile

<img width="539" alt="screen shot 2018-06-25 at 2 26 03 pm" src="https://user-images.githubusercontent.com/35348560/41869534-482759fc-7887-11e8-961d-fdcb19433c10.png">

--

### Date buckets
1. Design
Should now look like this:
<img width="278" alt="screen shot 2018-06-25 at 2 26 38 pm" src="https://user-images.githubusercontent.com/35348560/41869605-868812b8-7887-11e8-99a1-6169b3a51333.png">

- Box background colour + "X's": #222
- Text colour: #fff 
- Divider lines (1px) + "x circles (23px)": #ddd

2. Interaction
Make date selections (this box) area appear only when the first date has been selected.

--

### Calendar grey
<img width="487" alt="screen shot 2018-06-25 at 2 31 41 pm" src="https://user-images.githubusercontent.com/35348560/41869780-1a9b94c0-7888-11e8-89b9-32d2c65e78da.png">

1. Changing to all black at a regular weight, and bold the wed/thurs columns (including the day of the week titles).
2. Adding grey background (#eee) behind disabled dates

--

### On-state selection
Changing from squares to circles with a 50px hit-radius:
<img width="286" alt="screen shot 2018-06-25 at 2 58 20 pm" src="https://user-images.githubusercontent.com/35348560/41869819-4137e49e-7888-11e8-98eb-e8d26a416f68.png">

--

### Bolding
Only words on the page that should be bolded are now:
1. 3 days in the instructions
2. June 2018 (month + year) in the calendar
3.  days of the week in the calendar
4. "Your 3 selected days" in the date-selection box
5. the warning box messaging below the calendar

--

### Mobile
The calendar will take the full width, and dates selected will appear below it like this:
<img width="732" alt="screen shot 2018-06-25 at 3 05 50 pm" src="https://user-images.githubusercontent.com/35348560/41870173-4989783c-7889-11e8-903a-d0e3a11cd703.png">

emanelfy:
[...]

PR #235, opened 2018-07-05
Calendar Date Selection Redesign

nmakuch:
### Date Buckets
- Changing Date Bucket Design (Mobile and Desktop).
- Date Buckets are invisible until a date has been selected.
- Error Message design inside of buckets
- Mobile Styling of Buckets

| Before | After |
|--------|-------|
|  buckets-before|   buckets-after|

### Footer Glitch Fix
- Noticed a weird spacing in the Footer caused by the gcui element, removed it and created
- This glitch had been appearing on all pages in mobile view
- Updated the footer tests 
- Look is identical

### Calendar Redesign
- Disabled dates/weekday headings BG Gray, black text, regular weight
- Selectable days/weekday headings BG white, black text, bold weight
- Testing mobile Calendar on iOS and Android devices

| Before | After |
|--------|-------|
| calendar-before| calendar-after|

## Design Changes left out (Discussed with Eman):
- Going with squares instead of circles, we want the width of the calendar to be able to be liquid and this is easier with squares

# This PR will close [ISSUE 191](https://github.com/cds-snc/ircc-rescheduler/issues/191)


web/src/assets/CanadaWordmark.svg
```diff
@@ -0,0 +1,50 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Generator: Adobe Illustrator 22.1.0, SVG Export Plug-In . SVG Version: 6.00 Build 0)  -->
```
timarney: I would suggest running this SVG through 
https://jakearchibald.github.io/svgomg/

Should clean up all the unneeded Adobe comments etc... + will drop the files size a little (worth a try IMO)

web/src/assets/cancel.svg
```diff
@@ -1 +1,17 @@
-<svg id="Layer_1" data-name="Layer 1" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 22.63 22.63"><defs><style>.cls-1{fill:#fff;}.cls-2{stroke:#000;stroke-miterlimit:10;stroke-width:0.3px;}</style></defs><title>cancel</title><circle class="cls-1" cx="11.31" cy="11.31" r="10.81"/><path d="M16.11,16.64a.42.42,0,0,1-.3-.12L7.6,8.31a.41.41,0,0,1,0-.58.42.42,0,0,1,.59,0l8.21,8.21a.41.41,0,0,1,0,.58A.42.42,0,0,1,16.11,16.64Z" transform="translate(-0.69 -0.81)"/><path d="M7.89,16.64a.42.42,0,0,1-.29-.12.41.41,0,0,1,0-.58l8.21-8.21a.42.42,0,0,1,.59,0,.41.41,0,0,1,0,.58L8.19,16.52A.42.42,0,0,1,7.89,16.64Z" transform="translate(-0.69 -0.81)"/><path class="cls-2" d="M12,23.29A11.17,11.17,0,1,1,23.16,12.12,11.17,11.17,0,0,1,12,23.29Zm0-21.5A10.34,10.34,0,1,0,22.34,12.12,10.35,10.35,0,0,0,12,1.79Z" transform="translate(-0.69 -0.81)"/></svg>
```
timarney: Would suggest the same for all the SVGs https://jakearchibald.github.io/svgomg

web/src/components/ErrorMessage.js
```diff
@@ -53,7 +54,7 @@ class ErrorList extends React.Component {
 
   render() {
     return (
-      <div className={this.isEmpty() ? `empty ${noErrorList}` : errorList}>
```
timarney: Should be able to do something like {this.isEmpty ? null : errorList} here to save the css work ... might make it a little easier to see what's going on.  Also won't render a hidden div if it's not needed (not sure what is better on the a11y end - hidden item or item rendering an item after the initial load).

web/src/components/Calendar.js
```diff
@@ -343,40 +370,107 @@ const dayBox = css`
 `
 
 const daySelection = css`
+  background: ${theme.colour.blackLight} !important;
   margin-bottom: ${theme.spacing.xl};
+  width: 20rem !important;
+  padding: ${theme.spacing.lg} ${theme.spacing.lg} 0 ${theme.spacing.lg};
+
+  li {
+    border-top: 1px solid #dddddd;
```
pcraig3: Another colour here. Might end up being nuts trying to put them all into `styles.js`, so maybe let's hold off if this is the only place. But if we start re-using them, then yes.

web/src/components/Calendar.js
```diff
@@ -343,40 +370,107 @@ const dayBox = css`
...
+  height: 0;
+  margin-top: ${theme.spacing.lg};
+  margin-left: ${theme.spacing.md};
+  border-top: 20px solid transparent;
+  border-bottom: 20px solid transparent;
+  border-right: 25px solid ${theme.colour.blackLight};
+
+  ${mediaQuery.lg(css`
+    display: none;
+  `)};
```
pcraig3: This is super cool. Neat CSS trick. 👍

web/src/components/Calendar.js
```diff
@@ -550,12 +640,11 @@ class Calendar extends Component {
           onBlur={() => onBlur(value)}
           containerProps={{ id, tabIndex }}
         />
-        <div className={daySelection}>
-          <h3>
-            <Trans>Your 3 selected days:</Trans>
-          </h3>
+        <div className={value.length ? triangle : noDates} />
+        <div className={value.length ? daySelection : noDates}>
+          <h3>Your 3 selected days:</h3>
```
pcraig3: "Your 3 selected days:" should be inside of a `<Trans>` component.

web/src/components/Calendar.js
```diff
@@ -300,28 +331,24 @@ const calendarContainer = css`
...
+      font-size: ${theme.font.md};
     `)};
 
     &.empty {
-      background-color: ${theme.colour.greyLight};
-      border: 2px solid ${theme.colour.greyLight};
-
       * {
-        visibility: hidden;
+        display: none;
```
pcraig3: I was looking into this, whether setting an error message from `display:none` to `display:block` was a good idea.

I haven't found what looks like the definitive answer.

- [this article](http://web-accessibility.carnegiemuseums.org/code/alerts/) says to use `display:none` to hide error messages until we need them
- [this article](http://www.davidmacd.com/blog/test-aria-live-display-none.html) and [this one](https://developer.paciellogroup.com/blog/2012/06/html5-accessibility-chops-aria-rolealert-browser-support/) are both saying that it might not work on Safari

Let's keep your CSS but then we'll run this manually / see if the accessibility review picks this up.

web/src/components/ErrorMessage.js
pcraig3: i. the calendar component is looking for an `.empty` class to apply some CSS. Also, the test I wrote for this is looking for an `empty` class because I was trying to test that the logic was working properly.  Not sure it's the ideal solution, but for now would prefer to leave it in.

ii. Looked this up, and it looks like we want the empty div to be on the page when it loads.
- according to [some W3C aria guidance](https://www.w3.org/TR/WCAG20-TECHS/ARIA19.html?f5tp=1), it says "The error container must be present in the DOM on page load for the error message to be spoken by most screen readers."
- and then on [this scott vinkle blog post](https://medium.com/@svinkle/creating-accessible-react-apps-38dc1c2775cf), he says "Additionally, it’s important to always render the Announcement component in the template as some browser/screen reader technologies will not announce content when the aria-live element is dynamically added to the DOM. As a result, this component should always be included in any parent components in your app."

I was looking into whether `display:none` causes a problem here as well (see above comment), but I think that's probably okay.

web/src/components/ErrorMessage.js
```diff
@@ -86,7 +87,9 @@ class ValidationMessage extends React.Component {
   render() {
     return (
       <span
-        className={`${this.props.message ? '' : 'empty '}${errorMessage}`}
+        className={`${
+          this.props.message ? '' : `empty ${noError}`
+        }${errorMessage}`}
```
pcraig3: This could also be 
```
className={this.props.message ? errorMessage : `empty ${noError}`}
```

web/src/components/ErrorMessage.js
```diff
@@ -107,10 +110,19 @@ ValidationMessage.propTypes = {
...
+
+  isEmpty() {
+    return !(this.props.message ? true : false)
+  }
+
   render() {
     return (
       <span
-        className={`${this.props.message ? '' : 'empty '}${errorMessage}`}
+        className={this.isEmpty() ? `empty ${noError}` : errorMessage}
```
pcraig3: since `isEmpty` is super simple here, we could just use the check for the message.

Like this
```
 className={!(this.props.message) ? `empty ${noError}` : errorMessage}
```

And then we could remove the constructor and the `isEmpty` function.

pcraig3: Actually, it could be exactly the same as the comment before ☝️ 

```
className={this.props.message ? errorMessage : `empty ${noError}`}
```

web/src/components/Footer.js
```diff
@@ -116,12 +130,18 @@ const Footer = withI18n()(({ topBarBackground, i18n, context = {} }) => (
...
+          alt={
+            context.store.language === 'en'
+              ? 'Symbol of the Government of Canada'
+              : 'Symbole du gouvernement du Canada'
+          }
+          aria-label={
+            context.store.language === 'en'
+              ? 'Symbol of the Government of Canada'
+              : 'Symbole du gouvernement du Canada'
+          }
```
pcraig3: Pretty sure we only want one of these, either `alt` or `aria-label`. In this case, I think let's go with `alt`.

web/src/components/__tests__/Footer.test.js
```diff
@@ -38,4 +38,14 @@ describe('<Footer />', () => {
...
   })
+
+  it('renders with Canadawordmark in French', () => {
+    const footer = mount(<Footer context={getStore('fr')} />)
+    expect(footer.find('img').length).toBe(1)
+  })
+
+  it('renders with Canadawordmark in English', () => {
+    const footer = mount(<Footer context={getStore('en')} />)
+    expect(footer.find('img').length).toBe(1)
```
pcraig3: You should also get the "alt" attribute of the image and check that. There's definitely a way using enzyme, maybe [check here](https://github.com/airbnb/enzyme/issues/669#issuecomment-260086809). If you can't figure it out, let me know.

web/src/components/Calendar.js
pcraig3: Actually, I've just checked and we are using `#DBDBDB`, which is pretty much the same. Could swap this one for the one we already have.

PR #236, opened 2018-07-06
Refactor Calendar Tests to use non-hardcoded data

timarney:
This update uses the utility methods in calendarDates matching the Calendar output against dynamic data (start / end dates etc...).


timarney:
I think getting the no JS cal and potentially both cals setup to use an array of dates from getValidDays might end up being the way forward for moving the app into other locations.  This would keep the Wed, Thurs type of logic out of the Cal pages and just supply a list of dates to render.

PR #239, opened 2018-07-09
Fix some of the lingui translations we had for strings

Amirroracle:
So the essence of this ticket is that @pcraig3 fixed the issue of needing a translate method for when we need a string at runtime since the ```i18n``` object was null at times. So I can now just use ```i18n_()``` to get a translated string at runtime and this also means that ```yarn extract && yarn compile``` will generate the string in the designated json files.

### DIFF
--- a/web/src/components/__tests__/Reminder.test.js
+++ b/web/src/components/__tests__/Reminder.test.js
@@ -1,10 +1,16 @@
 import React from 'react'
 import { render } from 'enzyme'
 import Reminder from '../Reminder'
+import { I18nProvider } from 'lingui-react'
+import { i18n } from 'lingui-i18n/dist'
 
 describe('<Reminder />', () => {
   it('renders reminder message', () => {
-    const reminder = render(<Reminder>Message</Reminder>)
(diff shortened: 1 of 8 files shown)


web/src/components/__tests__/Reminder.test.js
```diff
@@ -1,10 +1,16 @@
...
+import { i18n } from 'lingui-i18n/dist'
 
 describe('<Reminder />', () => {
   it('renders reminder message', () => {
-    const reminder = render(<Reminder>Message</Reminder>)
+    const reminder = render(
+      <I18nProvider>
+        <Reminder i18n={i18n}>Message</Reminder>
+      </I18nProvider>,
+    )
```
pcraig3: I think we should be testing some of our content switches from English to French since we're using a Provider here and everything.

Alternatively, we could export our base components separately from our wrapped components so that we don't have to add i18n every time.

Amirroracle: Sounds good :+1: , I can create a new PR for this Thursday.

Issue #238, opened 2018-07-09
Field validation


timarney:
We're adding proper field validation to the form inputs:

Theres 2 cases we'll be looking to handle 

1. 2 many characters entered (we'll max at 2000)
2. Invalid Paper File #

These will be in addition of the you haven't entered any data

We'll need to figure out some updated wording to inform the user accordingly (no input vs bad input)

<img width="667" alt="screen shot 2018-07-09 at 11 27 52 am" src="https://user-images.githubusercontent.com/62242/42460311-8888e710-836b-11e8-82a7-84963e4decdf.png">

Note: For the text inputs there will also be HTML 5 validation in place to essentially block the user from entering too many characters.

:: @zakkain @emanelfy

pcraig3:
I put this in the Trello card, but I'll repeat it here for completeness.

Our service can only save 2000 characters (~300 words) so we need to stop people from entering super long explanations or names or whatever.

#### Full name: 
- cannot be empty
- cannot be more than {20 words | 200 characters}

#### Email:
- cannot be empty
- passes email regex + trim whitespace

#### Paper file number:
- cannot be empty
- always contains a letter, followed by 9 numbers (ie, C123456789)

#### Reason:
- cannot be empty
- check as an enum against known values

#### Explanation:
- cannot be empty
- cannot be more than {150 words | 1500 characters}

In addition to all this, we should automatically trim whitespace on all inputs. There is no case in which leading / trailing whitespace is significant that I can think of.

Feel free to weigh in, @emanelfy, @zakkain.

pcraig3:
I put this in the Trello card, but I'll repeat it here for completeness.

Our service can only save 2000 characters (~300 words) so we need to stop people from entering super long explanations or names or whatever.

#### Full name: 
- cannot be empty
- cannot be more than {20 words | 200 characters}

#### Email:
- cannot be empty
- passes email regex + trim whitespace

#### Paper file number:
- cannot be empty
- correctly formatted (_shhh_)

#### Reason:
- cannot be empty
- check as an enum against known values

#### Explanation:
- cannot be empty
- cannot be more than {150 words | 1500 characters}

In addition to all this, we should automatically trim whitespace on all inputs. There is no case in which leading / trailing whitespace is significant that I can think of.

Feel free to weigh in, @emanelfy, @zakkain.

zakkain:
**Full name**
`We're expecting a name that's shorter than 20 words. Please use the name recorded on your application.`

[...]

PR #241, opened 2018-07-11
Better Field Validation

timarney:
As per #238 this adds a new level of validation to the project.

Code has been setup to roughly drop / replace the existing validation.

validatorjs has been added which allows rules to be defined as simple objects

## Rules and Message Translation

```
let rules = {
  name: 'required',
  email: 'required|email',
  age: 'min:18'
};
```

Error messages can be customized for single elements which is what we're doing.

Typically that would be done as follows:
```
{
  'required.fullName': 'some message',
  'max.fullName': 'another message'
}
```

In our case given our translation setup the messages has been setup to be used as message keys

```
{
'required.fullName': 'fullNameErrorMessage',
}
```

An error messages object has been setup to handle the translations 

```
export const errorMessages = {}

errorMessages.fullNameErrorMessage = (
  <Trans>
    You need to tell us your name so we know who is requesting a new
    appointment.
  </Trans>
)
```

Allowing us to get error keys back and than display the translated message
```
errorMessages[this.props.message] i.e. errorMessages[fullNameErrorMessage] -> returns <Trans>The message</Trans>)

```

## Paper File Number Validation

The format describing this has been moved to a .env file

## Notes: 
- We'll need to update the translation files before merging this. 
- This validation will be used for the updated Node API as well

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -51,7 +51,8 @@
     "react-ga": "^2.5.3",
     "react-helmet": "^5.2.0",
     "react-router-dom": "^4.2.2",
-    "unfetch": "^3.0.0"
+    "unfetch": "^3.0.0",
+    "validatorjs": "^3.14.2"
   },
   "devDependencies": {
     "@types/graphql": "^0.13.0",
--- a/web/package.json
+++ b/web/package.json
@@ -89,6 +89,9 @@
       "^.+\\.css$": "<rootDir>/node_modules/razzle/config/jest/cssTransform.js",
       "^(?!.*\\.(js|jsx|css|json)$)": "<rootDir>/node_modules/razzle/config/jest/fileTransform.js"
     },
+    "globals": {
+      "RAZZLE_PAPER_FILE_NUMBER_PATTERN": "[a-zA-Z]{1}|[a-zA-Z]{1}" 
+    },
     "setupTestFrameworkScriptFile": "<rootDir>/test/setupTests.js",
     "testMatch": [
       "<rootDir>/**/__tests__/**/*.js",
--- a/web/package.json
+++ b/web/package.json
@@ -90,7 +90,7 @@
       "^(?!.*\\.(js|jsx|css|json)$)": "<rootDir>/node_modules/razzle/config/jest/fileTransform.js"
     },
     "globals": {
-      "RAZZLE_PAPER_FILE_NUMBER_PATTERN": "[a-zA-Z]{1}|[a-zA-Z]{1}" 
+      "RAZZLE_PAPER_FILE_NUMBER_PATTERN": "[a-zA-Z]{1}" 
     },
     "setupTestFrameworkScriptFile": "<rootDir>/test/setupTests.js",
     "testMatch": [
--- a/web/src/__tests__/Validation.tests.js
+++ b/web/src/__tests__/Validation.tests.js
@@ -25,9 +25,9 @@ it('Show correct error message when passing bad data', () => {
 
   expect(success).toEqual(false)
   expect(validate.errors.first('paperFileNumber')).toEqual(
-    'paperFileNumberErrorMessage',
+    'invalidPaperFileNumberErrorMessage',
   )
-  expect(validate.errors.first('email')).toEqual('emailErrorMessage')
+  expect(validate.errors.first('email')).toEqual('invalidEmailErrorMessage')
 })
 
 it('Show correct error message when passing invalid fields', () => {
--- a/web/src/__tests__/Validation.tests.js
+++ b/web/src/__tests__/Validation.tests.js
@@ -1,0 +1,126 @@
+import {
+  RegistrationFields,
+  CalendarFields,
+  getFieldNames,
+  defaultMessages,
+  paperFileNumberErrorMessage,
+  emailErrorMessage,
+  inErrorMessage,
+  selectedDaysEmptyErrorMessage,
+  selectedDaysMinMaxErrorMessage,
+  fullNameErrorMessage,
+  getFieldErrorStrings,
+} from '../validation'
+
+import Validator from 'validatorjs'
+
+it('Gets ar array of field names from Object Keys', () => {
+  expect(getFieldNames(RegistrationFields)[0]).toEqual('fullName')
+  expect(getFieldNames(CalendarFields)[0]).toEqual('selectedDays')
+})
+
+it('Show correct error message when passing invalid fields', () => {
+  const vals = {
+    fullName: 'John Li',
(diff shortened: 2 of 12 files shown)


zakkain:
Update from the issue discussion #238: 

Change from `Please be concise` in the Explanation error to `Please shorten your explanation`

web/src/validation.js
```diff
@@ -0,0 +1,175 @@
...
+  'required.fullName': 'fullNameErrorMessage',
+  'max.fullName': 'fullNameMaxErrorMessage',
+  'required.email': 'emailErrorMessage',
+  'email.email': 'invalidEmailErrorMessage',
+  'required.paperFileNumber': 'paperFileNumberErrorMessage',
+  'required.reason': 'reasonErrorMessage',
+  'required.explanation': 'explanationErrorMessage',
+  'max.explanation': 'explanationMaxErrorMessage',
+  'required.selectedDays': 'selectedDaysEmptyErrorMessage',
+  in: 'inErrorMessage',
```
pcraig3: Some of these are formatted differently.

We have:
1. {input name}ErrorMessage (ie, `fullNameErrorMessage`)
2. {input name}{qualifier}ErrorMessage (ie, `fullNameMaxErrorMessage`)
3. {qualifier}{input name}ErrorMessage (ie, `invalidEmailErrorMessage`)

I would suggest:
- we keep (1) and standardize either (2) or (3) (not bothered which)
- or, more radically, we toss (1) entirely and _always_ use (2) or (3)

web/src/validation.js
```diff
@@ -0,0 +1,175 @@
...
+    Sorry, there's a limit of 150 words for this explanation. Please be concise.
+  </Trans>
+)
+
+errorMessages.selectedDaysEmptyErrorMessage = (
+  <Trans>You must select 3 days. Please select 2 more days to continue.</Trans>
+)
+
+errorMessages.selectedDaysMinMaxErrorMessage = (
+  <Trans>Exactly three dates must be passed</Trans>
```
pcraig3: Also, just as a thing we have to keep in mind: none of these have french equivalents yet, so either when this goes it or directly afterwards, we should run `yarn extract` and get the translations in as well. (@anikbrazeau is probably the one to ask for translations for these error messages once @zakkain has given them a look over.)

web/src/validation.js
```diff
@@ -0,0 +1,175 @@
...
+}
+
+/*--------------------------------------------*
+ * Form Fields & Rules
+ *--------------------------------------------*/
+
+const getPaperFileNumberPattern = () => {
+  if (
+    !process.env.RAZZLE_PAPER_FILE_NUMBER_PATTERN &&
+    !typeof RAZZLE_PAPER_FILE_NUMBER_PATTERN
```
pcraig3: What does this second check do?

web/src/validation.js
```diff
@@ -0,0 +1,175 @@
...
+  ) {
+    // eslint-disable-next-line no-console
+    console.error('PAPER_FILE_NUMBER_PATTERN must be defined')
+    return null
```
pcraig3: Also, what do we think? If we have no environment variable, should we kill the app (like the API does)?

I tried running without the variable defined and it just always failed to validate my input, but the app was running okay.

web/package.json
```diff
@@ -88,6 +89,9 @@
       "^.+\\.css$": "<rootDir>/node_modules/razzle/config/jest/cssTransform.js",
       "^(?!.*\\.(js|jsx|css|json)$)": "<rootDir>/node_modules/razzle/config/jest/fileTransform.js"
     },
+    "globals": {
+      "RAZZLE_PAPER_FILE_NUMBER_PATTERN": "[a-zA-Z]{1}|[a-zA-Z]{1}"
```
pcraig3: We can take out this pipe symbol now that we're not using it in our .env file variable.

Or _are_ we? 🕵️ 

Even if we are (this is a secret), let's take it out.

web/src/__tests__/Validation.tests.js
```diff
@@ -0,0 +1,117 @@
...
+
+  const validate = new Validator(vals, RegistrationFields, defaultMessages)
+  const success = validate.passes()
+
+  expect(success).toEqual(false)
+  expect(validate.errors.first('paperFileNumber')).toEqual(
+    'invalidPaperFileNumberErrorMessage',
+  )
+  expect(validate.errors.first('email')).toEqual('invalidEmailErrorMessage')
+})
```
pcraig3: We should do a test or two for correctly validating stuff, like our email formatting.

Also trying with trailing whitespace and seeing that that works.

web/src/validation.js
```diff
@@ -0,0 +1,175 @@
...
+
+Validator.register(
+  'paper_file_number',
+  function(value, requirement, attribute) {
+    // eslint-disable-next-line security/detect-non-literal-regexp
+    const regex = new RegExp('^' + getPaperFileNumberPattern() + '$', 'i')
+    return regex.test(value)
+  },
+  'invalidPaperFileNumberErrorMessage',
+)
```
pcraig3: Should the email validation be in here somewhere as well? It looks like it was removed from the registration page.

web/src/validation.js
timarney: The second check is the JEST global as it doesn't have access to the .env

web/src/validation.js
timarney: Ya I'll throw an error here.

web/src/validation.js
timarney: Lib has email regex built in.

web/src/__tests__/Validation.tests.js
timarney: Tested an e-mail with whitespace and it will fail.

Thinking we should have a "clean inputs" function that gets run before passing to the validation.  

We'll want to cleaned inputs to pass to the api so thinking this should fall outside of the validator itself.  Maybe a new card would make sense for that case.

Also should we be updated the trimmed output back to the form?

web/src/validation.js
```diff
@@ -0,0 +1,172 @@
+import React from 'react'
+import Validator from 'validatorjs'
+import { Trans } from 'lingui-react'
+
+/*--------------------------------------------*
+ * Character limits 
+ *--------------------------------------------*/
+
+const inputFieldMaxChars = 500
+const textAreaMaxChars = 1500
```
Amirroracle: Sorry for the nit pick! But shouldn't these global constants be following a naming convention of UPPER_CASE?

timarney: 👍- updated

PR #243, opened 2018-07-11
Les dates ou sont-ils?

pcraig3:
This PR:

- adds back the `RAZZLE_PUBLIC_DIR` so that we can run `yarn start` locally again
- runs node with locale enabled in prod mode so that our noJS calendar page translates correctly
  - when we run `yarn dev`, we'll still get the wrong-looking dates, but since that never goes live we don't care. (so it won't solve [this problem](https://user-images.githubusercontent.com/2454380/42601979-89f8b8a2-8535-11e8-9301-b234b22db702.png))
- uses lowercase month names and day names for the JS calendar (this is what our formatting looks like everywhere else)

## Screenshots

| before | after |
|--------|-------|
|  month and weekday names capitalized, unlike selected day  | month and weekday names lowercased, matches selected day format      |
|  screen shot 2018-07-11 at 5 22 55 pm  |   screen shot 2018-07-11 at 5 26 00 pm    |
|  `yarn start` -> no js calendar not localised    |    `yarn start` -> no js calendar works as expected  |
|   screen shot 2018-07-11 at 5 22 06 pm   |   screen shot 2018-07-11 at 5 26 18 pm  |


web/src/server.js
```diff
@@ -28,7 +28,7 @@ server
   .use(helmet.noCache()) // Set Cache-Control, Surrogate-Control, Pragma, and Expires to no.
   .use(helmet.xssFilter()) // Sets "X-XSS-Protection: 1; mode=block".
   .disable('x-powered-by')
-  .use(express.static('./public'))
+  .use(express.static(process.env.RAZZLE_PUBLIC_DIR || './public'))
```
Amirroracle: will RAZZLE_PUBLIC_DIR have to be something we set ourselves in the .env file? It seems like it was there before but we removed it with regards to deployment issues, do you know if this will affect that?

pcraig3: Nope, this is already set by Razzle / After.js.

By removing it, we were just ignoring the value, but it's still getting set for us.

Amirroracle: 👍

PR #246, opened 2018-07-13
Remove band-aid solution for undefined i18n objects in the calendar and amend some tests from previous PR

Amirroracle:
For this PR, we firstly amend some tests that could use work from the previous pr, then remove the band-aid solution we came up with when the i18n object was null (when they refreshed the page). Now we should be able to resolve i18n strings at runtime without the fear of getting an ```i18n._ is undefined``` error. 

Would like some feedback on some of the tests that were changed.


pcraig3:
This is great!

PR #247, opened 2018-07-13
The app should be more fun

anikbrazeau:


pcraig3:
Hmm, I think we are not supposed to have fun here.

anikbrazeau:
Oh sorry my mistake. 😐

Issue #237, opened 2018-07-09
New page: Something went wrong (confirmation emails)


emanelfy:
Need messaging for a new "Something went wrong" page in the event that there's a system error and the email notification wasn't sent to staff and/or the user.

Should be along the lines of "something went wrong, please submit your request again" ... 

--
[@zakkain to submit copy for @pcraig3 @timarney]

timarney:
Here's the possible scenarios that I see:

1 - staff email failed
2 - user response email failed
3 - both staff and user email failed

We'll get back a fairly detailed response from Amazon (I have been able to track done a list showing possible errors as of yet)

There's also the possibility that an email failed but Amazon will retry sending X amount of times (I don't think we can handle this case as there's no time limit on this ... user has left the site).

zakkain:
@timarney so, this page would load _after_ the user hits "Send request" on the final page, and would essentially replace the `/confirmation` page that loads when everything works properly. Correct?

zakkain:
Elaborating on that: depending on which case Tim outlined above, we could present this as very different levels of severity. If the staff email failed, absolutely we should ask the user to repeat the process and try again.

But if the user confirmation failed while the staff email _succeeded_, then I would argue we shouldn't ask the user to repeat the process. We shouldn't fire a "Something went wrong" page—we should fire the confirmation page, but with an added piece explaining they won't get a confirmation email (and direct them/include the information to review, so they can write it down).

@emanelfy thoughts?

zakkain:
@emanelfy What do you think of this layout for **Case 1** - only user email fails?

![image](https://user-images.githubusercontent.com/546495/42648081-c4ae2e66-85d3-11e8-985b-450f66fb5d77.png)

obviously a rough mock, I made cutting and pasting screenshots, and not final copy. There's a much better way to handle this layout in design, have any thoughts?

emanelfy:
The error feels too buried on this page, could be missed. 
Needs to go above _"What happens next,"_ reversing the order? @zakkain 

_Please make a copy of your request_ -- also feels vague. Perhaps **"Please make note"** is a more action-oriented way of phrasing that, as in, you're definitely writing this down. (or taking a screenshot or whatever). 
I would also put this on a second line, so it doesn't appear to be such a long sentence and potentially could be missed when skimming.

I think we also need to make it super clear somehow that staff have a copy, it's just that users need to note it on their end. That everything else is okay and going according to plan.
[...]

PR #249, opened 2018-07-13
Node API + handle mail responses

timarney:
This PR moves the submission code / sending emails to work directly with the express server via a post  route.

- E-mail responses are now checked redirected per https://github.com/cds-snc/ircc-rescheduler/issues/237#issuecomment-404603417
- Confirmation page has been updated to handle a failed user email
- Logic for emails / templates was cleaned up a bit to get the code out of the route itself


web/src/email/email-template.js
```diff
@@ -14,37 +18,52 @@ function datesMarkup(dates, newline) {
...
+}
+/* eslint-enable security/detect-object-injection */
+
+let prefix = '../../../'
+
+if (
+  process.env.NODE_ENV === 'production' ||
+  process.env.NODE_ENV === 'development'
+) {
+  prefix += 'web'
```
pcraig3: This will mean that we can't do `yarn start` from the `web` directory, right?

Is that a use case we care about? For example, I've been running `yarn build && yarn start` from the `web directory most of the time to check what the 'prod' app looks like, but this change would make it hard to do that.

**However**

Since we're probably getting rid of the `web` directory anyway, **and** since we should be imitating prod as much as possible, you're probably doing the right thing.

web/src/email/email-template.js
```diff
@@ -14,37 +18,52 @@ function datesMarkup(dates, newline) {
...
+export function humanReadable(dates, locale = 'en') {
+  return dates.map(date => {
+    return format(date, 'dddd MMMM DD YYYY', {
+      locale: locales[locale],
+    })
+  })
+}
+/* eslint-enable security/detect-object-injection */
+
+let prefix = '../../../'
```
pcraig3: lol. 

this is just hard-coded because our path isn't going to change, right? Like we can safely leave this alone forever now?

web/src/email/email-template.js
```diff
@@ -14,37 +18,52 @@ function datesMarkup(dates, newline) {
...
 const readFileContent = async filename => {
-  const file = path.resolve(__dirname, `./email_templates/${filename}.html`)
-  const content = await readFile(file, 'utf-8')
+  const file = path.resolve(
+    __filename,
+    `${prefix}/email_templates/${filename}.html`,
+  )
+  const content = await readFile(file, 'utf-8').catch(e => {
+    return e.message
+  })
```
pcraig3: Is there a use-case where we might have an error in production based on some user input or something, or is this strictly our fault if the error gets triggered?

web/src/email/sendmail.js
```diff
@@ -0,0 +1,88 @@
...
+    throw new Error('AWS_SECRET_ACCESS_KEY was not found in the environment')
+  if (!process.env.AWS_ACCESS_KEY_ID)
+    throw new Error('AWS_ACCESS_KEY_ID was not found in the environment')
+  if (!process.env.IRCC_RECEIVING_ADDRESS)
+    throw new Error('IRCC_RECEIVING_ADDRESS was not found in the environment')
+  if (!process.env.SENDING_ADDRESS)
+    throw new Error('SENDING_ADDRESS was not found in the environment')
+  if (typeof process.env.SITE_URL === 'undefined')
+    throw new Error('SITE_URL was not found in the environment')
+}
```
pcraig3: Is this working for you locally, because I wouldn't expect these to get picked up.

Since we're not setting up `dotenv` ourself in the web folder, we've been relying on [Razzle's default env-finding configuration](https://github.com/jaredpalmer/razzle#build-time-variables).

> You can create your own custom build-time environment variables. They must start with RAZZLE_. Any other variables except the ones listed above will be ignored ...

Yeah, so is this working for you?

web/src/email/sendmail.js
```diff
@@ -0,0 +1,88 @@
...
+  let split = days.split(',')
+
+  const cleaned = split
+    .map(day => {
+      return day.trim()
+    })
+    .sort()
+
+  return cleaned
+}
```
pcraig3: - maybe we should sort the dates before we save them?
- if we're trimming values when they are saved, we shouldn't ever end up here.

**But**

Doesn't hurt to run this, so 👍

web/src/pages/ConfirmationPage.js
```diff
@@ -18,7 +21,66 @@ const contentClass = css`
...
+    display: block;
+  `)};
+
+  img {
+    ${mediaQuery.md(css`
+      float: left;
+      margin-top: ${theme.spacing.xs};
+      margin-right: ${theme.spacing.md};
+    `)};
+  }
```
pcraig3: Since this is pretty much the same as we're using in `LandingPage`, we could just put this in `components/Reminder.js` and export it separately from there.

web/src/pages/ConfirmationPage.js
```diff
@@ -18,7 +21,66 @@ const contentClass = css`
...
+`
+
+const Availability = styled('div')`
+  border-left: 2px solid #e1e1e1;
```
pcraig3: Does it make sense to use `greyLight` or `greyVeryLight` here?

web/src/queries.js
```diff
@@ -22,28 +22,3 @@ export const GET_USER_DATA = gql`
...
-        reason: $reason
-        paperFileNumber: $paperFileNumber
-        availability: $availability
-      }
-    ) {
-      requestId
-      messageId
-    }
-  }
-`
```
pcraig3: Oh whoops. Forgot to remove those others earlier.

Looks like we can just delete this whole file.

web/src/server.js
```diff
@@ -32,38 +42,62 @@ server
   .use(cookieParser(SECRET))
   .use(bodyParser.urlencoded({ extended: false }))
   .post('/submit', async (req, res) => {
+    let input = Object.assign({}, req.body) // make a new object
+    input.selectedDays = cleanDates(input.selectedDays)
+
+    const validateReg = new Validator(input, RegistrationFields)
+    const validateCal = new Validator(input, CalendarFields)
```
pcraig3: Might become annoying if we end up with lots of pages to validate things on, but for now this is pretty clean 👍

web/src/server.js
```diff
@@ -32,38 +42,62 @@ server
...
+
+    if (!validateReg.passes()) {
+      return res.redirect('/register')
```
pcraig3: Would these show any errors, or just send you back to the registration page?

web/src/routes.js
```diff
@@ -30,11 +30,17 @@ export default [
     exact: true,
     component: withProvider(ReviewPage),
   },
+  {
+    path: '/confirmation/:error',
+    exact: true,
+    component: withProvider(ConfirmationPage),
+  },
```
pcraig3: I was going to say that we could make confirmation not an exact path, but this is way more explicit.

web/src/server.js
```diff
@@ -32,38 +42,62 @@ server
...
+      if (
+        staffResponse.messageId === null &&
+        applicantResponse.messageId === null
+      ) {
+        // both emails failed to send
+        return res.redirect('/error')
+      }
 
-      /*
-      TODO: handle response gracefully
```
pcraig3: lol 👏

web/src/server.js
```diff
@@ -89,8 +123,7 @@ server
         ? res.redirect(res.locals.redirect)
         : res.send(html)
     } catch (error) {
-      // eslint-disable-next-line no-console
-      console.log(error.message, error.stack)
+      console.log(error.message, error.stack) // eslint-disable-line no-console
```
pcraig3: Should figure out how to send these to sentry, but for now 👍

pcraig3:
👇 also, our CircleCI stuff is failing because it can't run the API tests anymore.

The config for it is all in here: [.circleci](https://github.com/cds-snc/ircc-rescheduler/blob/master/.circleci/config.yml)

I think we want to:
- remove `api:` and everything under it
- remove `deploy_api:` and everything under it
- remove `- api` and `- deploy_api` from `workflows:` ([lines 106 -> 114](https://github.com/cds-snc/ircc-rescheduler/blob/master/.circleci/config.yml#L106-L114))

web/src/server.js
timarney: This just sends you back.  I was thinking of adding a param or something to run the validation if it exists on page load but, I can't think of a legitimate case where this should be happening.  From what I see it's a final pass.  If you've been through the form and haven't "messed" with anything throwing back to the page should never happen (unless I'm not seeing something).  So I skipped that handling for now.

web/package.json
```diff
@@ -90,7 +93,7 @@
       "^(?!.*\\.(js|jsx|css|json)$)": "<rootDir>/node_modules/razzle/config/jest/fileTransform.js"
     },
     "globals": {
-      "RAZZLE_PAPER_FILE_NUMBER_PATTERN": "[a-zA-Z]{1}" 
+      "RAZZLE_PAPER_FILE_NUMBER_PATTERN": "[a-zA-Z]{1}|[a-zA-Z]{1}"
```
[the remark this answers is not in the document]
timarney: Good catch missed that on the merge

web/src/__tests__/email-template.test.js
```diff
@@ -1,12 +1,12 @@
...
+import { buildParams } from '../email/email-template'
 
 const options = {
   htmlTemplate: '_test-rich',
   plainTemplate: '_test-plain',
   formValues: {
     fullName: 'John Li',
     paperFileNumber: '123456',
-    availability: ['2018-06-26', '2018-06-29', '2018-07-31'],
+    selectedDays: ['2018-06-26', '2018-06-29', '2018-07-31'],
```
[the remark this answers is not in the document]
timarney: ya makes it clearer IMO

web/src/server.js
timarney: Yes that's what I'm thinking.

web/src/__tests__/email.test.js
```diff
@@ -0,0 +1,79 @@
...
+})
+
+const options = {
+  htmlTemplate: '<div>Mail Test</div>',
+  plainTemplate: 'Mail Test',
+  formValues: {},
+  url: ' ',
+  receivingAddress: 'mock-sendmail@cds-snc.ca',
+  sendingAddress: 'receive-mock-sendmail@cds-snc.ca',
+}
```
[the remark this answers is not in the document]
timarney: I actually renamed / created a file (with the same name)

Those got moved to the email-template tests as that is part of building up the params / template vs sending the email

https://github.com/cds-snc/ircc-rescheduler/blob/de23341a88d3f594fe15869a83e6ef0880009b5b/web/src/__tests__/email-template.test.js

web/src/email/email-template.js
timarney: Was running into issues with path differences between the tests and dev.  Something we can revisit once all the dir updates are in place maybe.

web/src/email/email-template.js
timarney: Ya - should be left there now.  Cleaner solution would be nice.

web/src/email/email-template.js
timarney: This was triggering when I was moving files around so should be safe now.  Files exist and is in that dir.

web/src/email/sendmail.js
timarney: For SITE_URL or all.  The others did get picked up as I was sending emails to my personal email + the CDS account.

web/src/email/sendmail.js
timarney: I modified this a little for the whitespace PR  (move the sort out of the trim function) - will pick this up after we merge.

PR #252, opened 2018-07-13
Add bundle analyzer 

Amirroracle:
Added the bundle analyzer for our razzle-webpack config. This will allow us to see just how much resource each library and js file takes up. 

-Currently spits out a stats json and a report.html file in the ircc/web/build folder. 
-This can be run using ```yarn stats```
-You can even try running this on the production build by changing the ```stats``` command to use ```start``` instead of ```dev```, they give slightly different results. 
-This is only for the web folder, not sure about adding it to the api folder given our deployment strategy is changing, but it shouldn't be the most difficult thing to add in the future. 
-the bundle analyzer has different config options in of itself, can configure it to host a server instance instead of a static html report (which shows a bit more information like what the designated library size is when gzipped, parsed or not)

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -12,7 +12,7 @@
     "extract": "NODE_ENV=development lingui extract",
     "compile": "NODE_ENV=development lingui compile",
     "add-locale": "lingui add-locale", 
-    "stats" : "BUNDLE_CHECK=true yarn dev"
+    "stats": "mkdir -p reports && BUNDLE_CHECK=true yarn dev"
   },
   "dependencies": {
     "@cdssnc/gcui": "^0.0.30",
--- a/web/package.json
+++ b/web/package.json
@@ -12,7 +12,9 @@
     "extract": "NODE_ENV=development lingui extract",
     "compile": "NODE_ENV=development lingui compile",
     "add-locale": "lingui add-locale", 
-    "stats": "mkdir -p reports && BUNDLE_CHECK=true yarn dev"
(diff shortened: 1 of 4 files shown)


timarney:
- Currently spits out a stats json and a report.html file in the ircc/web/build folder.
Wondering if the report should be output to a reports dir and also Git ignored. As far as I see we'll only be doing local analysis.  Just wondering about security.

- "This is only for the web folder" -> No worries on this as the API dir will be removed once the Node API PR gets merged.

web/package.json
```diff
@@ -11,7 +11,8 @@
     "test": "node --icu-data-dir=./node_modules/full-icu node_modules/jest/bin/jest.js",
     "extract": "NODE_ENV=development lingui extract",
     "compile": "NODE_ENV=development lingui compile",
-    "add-locale": "lingui add-locale"
+    "add-locale": "lingui add-locale", 
+    "stats" : "BUNDLE_CHECK=true yarn dev"
```
timarney: Tiny nit  "stats" : 

// remove the space

timarney:
Also of note to go with this

We can run 

```
yarn why package_name_here
```

To help track down dependancies

screen shot 2018-07-16 at 9 45 47 am


https://yarnpkg.com/lang/en/docs/cli/why/

web/package.json
Amirroracle: good catch, surprised Prettier didn't adjust it.

PR #255, opened 2018-07-16
Clean whitespace from inputs

timarney:
1. Cleans whitespace from inputs pre-validation

```
static validate(values) {
    const validate = new Validator(
     trimInput(values),
      ... )
```

2. Saves the trimmed values 

```
val = _whitelist({ val, fields })
val = trimInput(val)
```

3. Not in this PR as we should wait on https://github.com/cds-snc/ircc-rescheduler/pull/249 but will run trimInput() prior to sending to the Mailer.

Note: As part of this I also upgraded Final Form to use the latest.


web/src/pages/CalendarPage.js
```diff
@@ -139,7 +139,11 @@ class CalendarPage extends Component {
   }
 
   static validate(values) {
-    const validate = new Validator(values, CalendarFields, defaultMessages)
+    const validate = new Validator(
+      trimInput(values),
```
pcraig3: Ah, right, I forgot we would have to do this.

This is a bit strange because it means that we will always be validating against not the real values. 
Since this logic is also in the withProvider function below, we should be okay here.

But if we start getting more requirements around this stuff, we'll probably have to refactor it so that we can keep all the validation + cleaning stuff together.

timarney: Agreed

PR #257, opened 2018-07-16
Fix up the bundle anayzer to generate two different files in the same directory

Amirroracle:
-Last bundle analyzer PR had some issues with it, the prod command didn't work without yarn build. and one report was generated in the right place and the other one wasn't. So in this PR I am only adding one command to the package.json file instead of previously two, so we should only need to see the prod version of the bundle. This is also way cleaner since this means a server doesn't need to run to generate the reports.

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -12,8 +12,7 @@
     "extract": "NODE_ENV=development lingui extract",
     "compile": "NODE_ENV=development lingui compile",
     "add-locale": "lingui add-locale", 
-    "stats": "mkdir -p reports && BUNDLE_CHECK=true yarn dev",
-    "stats-prod": "mkdir -p reports && BUNDLE_CHECK=true yarn start"
+    "stats": "mkdir -p reports && BUNDLE_CHECK=true yarn build"
 
   },
   "dependencies": {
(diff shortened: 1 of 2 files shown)


web/package.json
```diff
@@ -12,9 +12,7 @@
     "extract": "NODE_ENV=development lingui extract",
     "compile": "NODE_ENV=development lingui compile",
     "add-locale": "lingui add-locale", 
-    "stats": "mkdir -p reports && BUNDLE_CHECK=true yarn dev",
-    "stats-prod": "mkdir -p reports && BUNDLE_CHECK=true yarn start"
-
+    "stats": "mkdir -p reports && BUNDLE_CHECK=true yarn build"
```
pcraig3: Looks like this creates the directory just fine even without the `mkdir` command. So you could even trim this down to `BUNDLE_CHECK=true yarn build`.

Amirroracle: I believe it would work without the mkdir command only if you already have a reports folder existing. Otherwise it'll throw an error.

pcraig3: Hey, I ran this locally and I got @nmakuch to run it as well. 

Seems like it will create the folder with or without the `mkdir` command, so I've removed it. 

If you run into problems, we can put it back.

Amirroracle: Ok cool np.

PR #258, opened 2018-07-17
Cal validation

timarney:
Fixes Calendar validation regression to display appropriate error message based on the number of dates selected.

i.e. You must select 3 days. Please select 1 more day to continue. 

The error messages use a combination of validatorjs + some logic within the Form page to add additional information about the # of dates selected.

validatorjs uses true/false to determine an error an expects a string back for the message so there isn't a way to customize based on a condition.  I was going to add some additional validators for different counts but adding the logic into the Form seems to be the cleaner option.

### DIFF
--- a/web/src/pages/CalendarPage.js
+++ b/web/src/pages/CalendarPage.js
@@ -207,45 +207,72 @@ class CalendarPage extends Component {
             values,
             errors,
             submitError,
-          }) => (
-            <form onSubmit={handleSubmit} className={fullWidth}>
-              <div>
-                <div
-                  id="submit-error"
-                  tabIndex="-1"
-                  className={focusRing}
-                  ref={errorContainer => {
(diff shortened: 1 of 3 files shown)


web/src/pages/CalendarPage.js
```diff
@@ -207,45 +207,72 @@ class CalendarPage extends Component {
...
+                    }}
+                  >
+                    <ErrorList message={err || ''}>
+                      {Object.keys(this.validate(values)).map((formId, i) => (
+                        <HashLink to={`#${formId}`} key={i}>
+                          {labelNames(formId) ? labelNames(formId) : formId}
+                          <br />
+                        </HashLink>
+                      ))}
+                    </ErrorList>
```
pcraig3: This has knock-on effects for @nmakuch's PR #266, but let's resolve those later.

PR #260, opened 2018-07-17
Better aria-label for "Remove day" button

pcraig3:
**Note: the only real changes in here are in `Calendar.js` and `Calendar.test.js`. The rest are all changes to the lingui stuff so that I could add the phrase "Remove day"**

***

## Remove day button alt text works in english and french 

Both the "remove day" text and the date itself will be translated when the app is switched into french.

Note that the alt-text for the images isn't having the date added to it since there's pretty much no way to focus on the image instead of the button.

Adding the alt-text means that it will pass the automated accessibility scanning software audits (running axe, for example), but the aria-label of the button is the thing screenreaders will end up on.

Kept most of the i18n stuff outside of the the renderDayBoxes function since it really just needs some static text.

Also added some tests.

### gif

aria-label


timarney:
Looks good.

PR #266, opened 2018-07-17
New Calendar Design

nmakuch:
# This PR Consists of:

## Calendar Bucket Restyling:
| Before | After |
|--------|-------|
|    screen shot 2018-07-17 at 11 15 31    |   screen shot 2018-07-17 at 11 16 54    |

## Updated data uri's for the arrows in the calendar to chevrons:

| Before | After |
|--------|-------|
 |   screen shot 2018-07-17 at 13 48 25 |  screen shot 2018-07-17 at 13 48 12  |

## Calendar dates are more square-like on major screens for desktop:

| Before  | After |
|--------|-------|
|    screen shot 2018-07-17 at 13 53 58    |    screen shot 2018-07-17 at 13 55 57   |

## Calendar weeks have more balanced padding

| Before | After |
|--------|-------|
|   screen shot 2018-07-17 at 14 17 26     |  screen shot 2018-07-17 at 14 04 46     |
- Noted a bug occurring on Safari and Firefox (as seen in 'Before' , the days break out of the container) in this PR.
  - Basically, DayPicker-Month needs to be just a bit smaller than the width of the container or it will break.

### Chrome Browser

| Before | After |
|--------|-------|
|  screen shot 2018-07-17 at 14 55 45   | screen shot 2018-07-17 at 14 55 09 |

### Firefox Browser

| Before | After |
|--------|-------|
|  screen shot 2018-07-17 at 15 00 06 |  screen shot 2018-07-17 at 15 00 29 |

### Safari Browser

| Before | After |
|--------|-------|
|  screen shot 2018-07-17 at 15 05 26   |  screen shot 2018-07-17 at 15 05 57  |

## Removing ErrorList:

| Before | After |
|--------|-------|
|  screen shot 2018-07-17 at 17 19 44 screen shot 2018-07-17 at 17 19 32  |  screen shot 2018-07-17 at 17 20 48 screen shot 2018-07-17 at 17 20 02  |
- Removed ErrorList from the JS side of the calendar
- Created ErrorCalendar in ErrorMessage, the submitError and selectedDaysError are now using this. 
  - Would be nice to get these to trigger one instance of this instead of two


timarney:
Nothing this as discussed will be fixed via a new issue.  

 screen shot 2018-07-20 at 9 25 30 am

Case:
- Select 0 days -> submit.
- Select more then 3 days

web/src/components/Calendar.js
```diff
@@ -272,18 +282,24 @@ const calendarContainer = css`
...
-    width: 100%;
-    display: block;
-    > div:first-of-type,
+  ${mediaQuery.sm(css`
+    > div:first-of-type {
+      width: 100%;
+    }
     > div:last-of-type {
       width: 100%;
     }
```
pcraig3: Same as the last one ☝️ 

These rules could be grouped together.

web/src/components/Calendar.js
```diff
@@ -338,34 +353,24 @@ const dayBox = css`
 `
 
 const daySelection = css`
-  background: ${theme.colour.blackLight};
+  background: #e1e1e1;
```
pcraig3: `greyLight` is nearly the same as this. If the contrast flag passes, we should tend towards colours we're already using.

web/src/components/Calendar.js
```diff
@@ -384,19 +388,25 @@ const daySelection = css`
       margin-bottom: ${theme.spacing.xs};
     }
 
+    li:last-of-type {
+      padding-bottom: ${theme.spacing.lg};
+      padding-bottom: ${theme.spacing.md};
```
pcraig3: These are the same rule! 🤓 🤓

web/src/components/Calendar.js
```diff
@@ -472,12 +462,11 @@ const renderDayBoxes = ({
...
-            <div className={removeDateDesktop}>
-              <img src={Cancel} alt={removeDayAltText} />
-            </div>
-
-            <div className={removeDateMobile}>
-              <img src={MobileCancel} alt={removeDayAltText} />
+            <div className={removeDate}>
+              <img
+                src={MobileCancel}
+                alt={locale === 'en' ? 'Remove day' : 'Supprimer cette journée'}
```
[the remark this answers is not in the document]
pcraig3: Actually, if we use `removeDayAltText` here, we should be fine.

web/src/components/ErrorMessage.js
```diff
@@ -131,4 +153,25 @@ ErrorMessage.propTypes = {
...
+    )
+  }
+}
+
+ErrorCalendar.propTypes = {
+  message: PropTypes.oneOfType([PropTypes.string, PropTypes.element]),
+  id: PropTypes.string,
+}
+
+export { ErrorMessage as default, ValidationMessage, ErrorList, ErrorCalendar }
```
pcraig3: This new error is almost identical to `ErrorMessage`. 
The two differences are:
- that one is a span and one is a div
- the CSS rules

I think we are (_were_) only using ErrorMessage on the calendar page, so we can change it however we like.

PR #267, opened 2018-07-17
Only install sentry in production mode

pcraig3:
To run sentry locally now, you're going to have to run `yarn start`

Notes:
- could have also set the DSN string to empty, as per this comment
  https://github.com/getsentry/raven-js/issues/105#issuecomment-340215644
  but not installing at all seems cleaner


timarney:
We should handle this case even though we won't hit it for dev.  

Error page "tell us what went wrong" link

screen shot 2018-07-18 at 8 50 14 am

PR #275, opened 2018-07-18
Landing Page Icon

nmakuch:
# This PR consists of:

## Changing language/Adding Icon on the Day Limit prompt found on the Landing Page

| Before | After |
|--------|-------|
|    screen shot 2018-07-18 at 16 12 23    |   screen shot 2018-07-18 at 16 12 17|


web/src/pages/LandingPage.js
```diff
@@ -21,6 +21,39 @@ const contentClass = css`
...
...uMzIsMi40MiwyLjQsMi4zNWExLjQsMS40LDAsMCwwLDEuNTQtMS41OUMxNy43Myw4LjE1LDE3LjcyLDcuMzksMTcuNzIsNi42MlpNNDAsNi41MnYyYzAsMS42LjM2LDEuOTUsMiwxLjk1aC4zMWMxLjEzLDAsMS42MS0uNDksMS42Mi0xLjYyLDAtMS4yOCwwLTIuNTcsMC0zLjg1LjExLTIuNTctLjc3LTIuNDQtMi41OS0yLjRBMS4yNywxLjI3LDAsMCwwLDQwLjA1LDRDNDAsNC44Niw0MCw1LjY5LDQwLDYuNTJaIi8+PHBhdGggZD0iTTQ4LjU3LDMxLjUxQTE0LjQyLDE0LjQyLDAsMSwxLDM0LjEzLDQ1Ljk0LDE0LjQ2LDE0LjQ2LDAsMCwxLDQ4LjU3LDMxLjUxWk0zNi43Nyw0NS45QTExLjc5LDExLjc5LDAsMSwwLDQ4LjU5LDM0LjEyLDExLjc5LDExLjc5LDAsMCwwLDM2Ljc3LDQ1LjlaIi8+PHBhdGggZD0iTTE1Ljc1LDI4Ljg2YTIuNjIsMi42MiwwLDEsMS0yLjYtMi42MkEyLjYyLDIuNjIsMCwwLDEsMTUuNzUsMjguODZaIi8+PHBhdGggZD0iTTMxLjUsMjguODlhMi42NCwyLjY0LDAsMCwxLTIuNjEsMi42MSwyLjYzLDIuNjMsMCwwLDEsMC01LjI2QTIuNjQsMi42NCwwLDAsMSwzMS41LDI4Ljg5WiIvPjxwYXRoIGQ9Ik0xNS43NSw0My4zMmEyLjYyLDIuNjIsMCwxLDEtMi42NC0yLjYzQTIuNTksMi41OSwwLDAsMSwxNS43NSw0My4zMloiLz48cGF0aCBkPSJNNDYuMzgsNDcuNzZsNC4yOC00LjMzYy42NC0uNjQsMS4yOC0xLjI5LDEuOTMtMS45MmExLjMyLDEuMzIsMCwxLDEsMS44NywxLjg0cS0zLjQ5LDMuNTEtNyw3YTEuMzcsMS4zNywwLDAsMS0yLjE3LDBjLS44OC0uODctMS43Ni0xLjc0LTIuNjMtMi42M2ExLjMzLDEuMzMsMCwwLDEtLjExLTIsMS4zMiwxLjMyLDAsMCwxLDIsLjEzQzQ1LjEyLDQ2LjQ2LDQ1LjcsNDcuMDcsNDYuMzgsNDcuNzZaIi8+PC9nPjwvZz48L3N2Zz4=');
```
pcraig3: https://www.w3.org/WAI/tutorials/images/decorative/

> Decorative images don’t add information to the content of a page. For example, the information provided by the image might already be given using adjacent text, or the image might be included to make the website more visually attractive.

Our calendar icon is strictly a visual cue, right?

I think it is: if people are reading the text, they will get the sense of what they are going to have to do. If they focused with a screenreader, the message "calendar icon" isn't really super useful, I don't think.

Devil's advocate: we could have it say "use the calendar to select three dates between august and october" or something.

pcraig3: Thoughts: @zakkain, @emanelfy, @nmakuch?

web/src/pages/LandingPage.js
```diff
@@ -85,18 +118,20 @@ class LandingPage extends React.Component {
...
+          <div className={messageContainer}>
+            <div className={iconContainer}>
+              <CalendarIcon />
+            </div>
+            <p>
+              <Trans>Then you’ll select</Trans>{' '}
+              <strong>
+                <Trans>3 days</Trans>
+              </strong>{' '}
+              <Trans>you&#39;re available between </Trans>{' '}
```
pcraig3: Since we're changing this around, we have to do another `yarn extract`, `yarn compile` otherwise the french doesn't translate.

image

Let's wait until just before we're about to merge to do that though.

web/src/pages/LandingPage.js
```diff
@@ -85,18 +118,20 @@ class LandingPage extends React.Component {
...
+              {getStartMonthName(new Date(), locale)} <Trans>and</Trans>{' '}
+              {getEndMonthName(new Date(), locale)}{' '}
```
pcraig3: Should end with a period, and we don't need the extra whitespace.

So this line should be `{getEndMonthName(new Date(), locale)}.`

web/src/pages/LandingPage.js
pcraig3: If we _do_ want alt-text, then we can't use CSS -- we have to make this an `` tag or something.

zakkain: The purpose of this icon is augment copy already on the page. In that sense, it is not strictly decorative but actually has semantic value as an icon 

that’s my take

zakkain: The purpose of this icon is augment copy already on the page. In that sense, it is not strictly decorative but actually has semantic value as an icon 

that’s my take

zakkain: that being said, the icon does not stand on its own. If we did have all text, it would just be something like “an icon of a calendar “

pcraig3: @zakkain, for visual users it's going to augment the copy necessarily, but screenreaders (ie, anyone who isn't looking at the page), it won't be there at all. 

If we want it to be picked up, we would want it to make sense when read out to someone, keeping in mind that it would be read _first_ before the rest of the sentence was read out.

zakkain: ah ok. I think that is A-OK. “A calendar showing 3 dates highlighted” ?

web/src/pages/LandingPage.js
pcraig3: Also let's see if we can make this a curly quote for `you’re`. Maybe `&rsquo;`?

web/src/pages/LandingPage.js
pcraig3: Final decision: no alt-text. 

Means it's fine to leave it as a CSS background image.

PR #277, opened 2018-07-19
Adds Sentry error logging for failed AWS emails

timarney:
- Previously errors we're logged in the console without notice of failure being reported back (outside of redirecting to an error page)

- Added the Node version of Raven and a call to capture the exception

- Moving forward we should move the Raven credentials into an .env file so we can re-use in multiple spots.

## Testing 

The easiest way I have found to test so far is to adjust the Params in the server code to use a null sending address

```
const applicantOptions = {
    htmlTemplate: 'applicant-rich',
    plainTemplate: 'applicant-plain',
    ...
    sendingAddress:null,
  }
```

### Reported Error

screen shot 2018-07-19 at 11 36 32 am

### DIFF
--- a/web/src/server.js
+++ b/web/src/server.js
@@ -16,6 +16,10 @@ import {
   cleanDates,
   sendMail,
 } from './email/sendmail'
+import Raven from 'raven'
+Raven.config(
+  'https://a2315885b9c3429a918336c1324afa4a@sentry.io/1241616',
+).install()
 
 // eslint-disable-next-line security/detect-non-literal-require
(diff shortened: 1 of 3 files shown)


web/src/server.js
```diff
@@ -16,6 +16,10 @@ import {
   cleanDates,
   sendMail,
 } from './email/sendmail'
+import Raven from 'raven'
```
pcraig3: Ah, so `raven` is different from `raven-js`?

timarney: Ya tracked down a project called raven-node which has now been rolled into the main repo.

PR #280, opened 2018-07-20
Add sourcemaps for sentry logging

Amirroracle:
Use the sentry webpack plugin to update our artifacts in sentry with the proper sourcemaps.

Notes to keep in mind:
-Running this requires an api key with ```project:write selected under scopes.``` under the sentry ui when you generate one. 
https://docs.sentry.io/clients/javascript/sourcemaps/
-The api key is an environment variable

### DIFF
--- a/web/razzle.config.js
+++ b/web/razzle.config.js
@@ -25,6 +25,19 @@ module.exports = {
         }),
       )
     }
+    if (process.env.RAZZLE_SENTRY_API) {
+      const SentryPlugin = require('@sentry/webpack-plugin')
+      config.plugins.push(
+        new SentryPlugin({
+          organization: 'canadian-digital-service',
+          project: 'ircc-rescheduler',
+          release: 'a2315885b9c3429a918336c1324afa4a',
+          include: '.',
+          ignore: ['node_modules', 'webpack.config.js'],
+          apiKey: process.env.RAZZLE_SENTRY_API,
+        }),
+      )
+    }
     return config
   },
 }
--- a/web/src/server.js
+++ b/web/src/server.js
@@ -17,9 +17,21 @@ import {
   sendMail,
 } from './email/sendmail'
 import Raven from 'raven'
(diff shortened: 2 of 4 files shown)


web/razzle.config.js
```diff
@@ -25,6 +25,19 @@ module.exports = {
         }),
       )
     }
+    if (process.env.RAZZLE_SENTRY_API) {
+      const SentryPlugin = require('@sentry/webpack-plugin')
+      config.plugins.push(
+        new SentryPlugin({
+          organization: 'canadian-digital-service',
+          project: 'ircc-rescheduler',
+          release: 'a2315885b9c3429a918336c1324afa4a',
```
timarney: Should this be dynamic?  What exactly does it do?

web/src/server.js
```diff
@@ -17,9 +17,21 @@ import {
   sendMail,
 } from './email/sendmail'
 import Raven from 'raven'
-Raven.config(
-  'https://a2315885b9c3429a918336c1324afa4a@sentry.io/1241616',
-).install()
+Raven.config('https://a2315885b9c3429a918336c1324afa4a@sentry.io/1241616', {
```
timarney: Thinking we need to do this client side as well as this is the Node setup.

web/razzle.config.js
Amirroracle: Organization and project I've made to be the exact strings we use in Sentry and release is the "release" version we send in ```server.js``` and ```Document.js``` so if we make it dynamic here we should make it dynamic there too.

web/src/server.js
Amirroracle: I thought that too, but the node docs says to add it and js docs doesn't specify.

timarney: As discussed this seems to do the trick as is.

timarney:
Going to approve and we'll get on staging for further testing.

PR #281, opened 2018-07-23
GA tracking for users with no JavaScript

timarney:
- Uses background image to fire a get request to google-analytic.com/collect
- Will count a pageview for "/nojs" page (page url must be encoded param)
- comments in code note url for details and required params

**Tested using a personal Google Analytics account**
no-js-view

**Details here on tracking setup (parameters)**
https://developers.google.com/analytics/devguides/collection/protocol/v1/devguide#proxy-server-tracking


web/src/Document.js
```diff
@@ -29,6 +51,9 @@ class Document extends React.Component {
           {helmet.link.toComponent()}
         </head>
         <body {...bodyAttrs}>
+          <noscript>
+            <div dangerouslySetInnerHTML={noJSTracker()} />
+          </noscript>
```
pcraig3: This nests a `<div>` inside of a `<div>`. Probably better to keep the logic somewhere else so no problem.

web/src/Document.js
```diff
@@ -1,6 +1,28 @@
...
+Required Params:
+v=1              // Version.
+&tid=UA-XXXXX-Y  // Tracking ID / Property ID.
+&cid=555         // Anonymous Client ID.
+&t=              // Hit Type.
+*/
+
+const noJSTracker = () => {
+  const uri = 'https://www.google-analytics.com/collect?v=1&cid=5555'
+  const ga = process.env.RAZZLE_GA_ID
```
pcraig3: If this variable isn't set, the `<noscript>` tag will still be inserted into the page. (The pageview doesn't go anywhere, so practically it's not a big deal.)

Would it be possible to return `<noscript><div style ... ></div></noscript>` and then do check for the variable existing, or is this hard to do in Document.js because we have to use `dangerouslySetInnerHTML`?

web/src/Document.js
```diff
@@ -1,6 +1,28 @@
...
+  const page = '%2Fnojs'
+  return {
+    __html: `
+    <div style="background-image: url('${uri}&t=pageview&dp=${page}&tid=${ga}')"></div>`,
```
pcraig3: `It looks like `page` is always going to be `no-js` here, whatever the URL actually is. 
I was thinking we would send something back like `/?nojs=true` or `/register?nojs=true`.

Is that hard to do / worth doing?

My hesitation is that, since our client IDs are always the same and our URLs are always the same, we wouldn't be able to tell if 5 `/nojs` results represent 5 people hitting the start page or 1 person going through the flow.

web/src/Document.js
```diff
@@ -1,6 +1,28 @@
...
+
+Required Params:
+v=1              // Version.
+&tid=UA-XXXXX-Y  // Tracking ID / Property ID.
+&cid=555         // Anonymous Client ID.
+&t=              // Hit Type.
+*/
+
+const noJSTracker = () => {
+  const uri = 'https://www.google-analytics.com/collect?v=1&cid=5555'
```
pcraig3: There are 4 `5`s here, but the comment above uses 3 of them. Did you use too many?

web/src/Document.js
timarney: Good point we should be able to pick up the real page name here.

web/src/Document.js
timarney: Thinking I can check for the existence of the variable and than insert the noscript tag.

web/src/Document.js
timarney: ya too many

web/src/Document.js
timarney: Originally I was trying to apply this as a css style to the body.  Just tested applying direct to the noscript tag with a single div inside and that seems to work.

web/src/Document.js
```diff
@@ -1,6 +1,30 @@
...
+v=1              // Version.
+&tid=UA-XXXXX-Y  // Tracking ID / Property ID.
+&cid=555         // Anonymous Client ID.
+&t=              // Hit Type.
+*/
+
+const noJSTracker = (path = '/') => {
+  const uri = 'https://www.google-analytics.com/collect?v=1&cid=555'
+  const ga = process.env.RAZZLE_GA_ID
+  const hasParams = () => (path.indexOf('?') === -1 ? '?' : '&')
```
pcraig3: Oh right, I forgot we sometimes have query variables. 

Although, since this url is only going to google, we could just cut them all off and send the `?nojs=true` as the only query variable we set. Doesn't hurt to send them though.

web/src/Document.js
```diff
@@ -1,6 +1,30 @@
...
+  const page = encodeURIComponent(path + hasParams() + 'nojs=true')
```
pcraig3: Since we're just sending this URL to Google, we could cut off other query params if we wanted to (but leaving them in gets us slightly more information potentially).

👍 from me.

PR #283, opened 2018-07-23
Refactor ErrorMessage styling + logic (ie, remove ErrorCalendar)

pcraig3:
**Note that there's still an issue with the calendar page submit error being out of sync with the values on the calendar itself. this is a known issue and causes a bit of a weird interaction, but this PR isn't meant to solve it.**

***

This PR cleans up a few things after #266 went in. It's not fixing everything but it improves things.

- consolidates some duplicated CSS rules
- uses our theme colours (`#DBDBDB` instead of `#E1E1E1`) for selected days list
- removes the hard-coded alt text for the "Remove day" buttons
- removes `ErrorCalendar` in favour of `ErrorMessage`
  - updates the `ErrorMessage` styling to match what `ErrorCalendar` was doing
- improves some logic bugs we had going on:
  - previously, the "you must select 3 days message" would never leave the page once it was triggered
  - previously, no error message at all would appear if you immediately pressed "Review request" on the calendar page

Updated ErrorMessage tests to account for new wrapping element (`div` instead of a `span` now), but that's all I did in terms of updating tests.

# Screens

| before | after |
|--------|-------|
|   error message would never disappear   |  error message will (eventually) disappear   |
|  message-2   |  message-1  |
|  error doesn't appear if button pressed immediately | error appears as normal now  |
|message-3 | message-4 |


timarney:
This is great.  One thing I noticed is this fixes the issue where the error would never leave the page there's an extra click between the error leaving and the "you can only select X dates" to show up error.

Maybe we tackle that under a separate PR

clicks

PR #286, opened 2018-07-24
Don't swap error messages on change

timarney:
By default Final Form was validates inputed data as the user types.  This update stores the previous error messages from the last submission and returns them onChange.  Waiting for the next submit to update the messages again.


timarney:
Nice catch.  Adding the windowExists seems to fix the issue.

PR #288, opened 2018-07-24
Update calendar tests to use valid days arrays for tests

timarney:
Update calendar tests to use valid days arrays for tests

This updates adds a new function to seek out a month to use for the tests that has at least 3 valid dates.

The tests have been fragile as we have been relying either upon the inital view to have multiple dates available or simulating clicks on the view to get to a month where we have enough dates.  This was falling out of sync with the dates trying to be selected.

### DIFF
--- a/web/src/components/__tests__/Calendar.test.js
+++ b/web/src/components/__tests__/Calendar.test.js
@@ -12,11 +12,6 @@ import {
 import parse from 'date-fns/parse'
 import addMonths from 'date-fns/add_months'
 import format from 'date-fns/format'
-import startOfMonth from 'date-fns/start_of_month'
-
-const firstDayNextMonth = () => {
-  return startOfMonth(addMonths(new Date(), 1))
-}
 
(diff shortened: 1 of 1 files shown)


pcraig3:
Does this need to be rebased or something after #287 goes in, or what were you thinking?

timarney:
Correct will rebase this and go from there.

web/src/components/__tests__/Calendar.test.js
```diff
@@ -58,10 +53,41 @@ const dayMonthYear = date => {
...
+  return accumulator
+}
+
+// find a month that has at least 3 dates we can use for tests
+const useMonth = dates => {
+  const groupedDates = dates.reduce(collectDates, {})
+
+  let month = false
+
+  Object.keys(groupedDates).forEach(function(key) {
```
pcraig3: could make this an es6 function but 🤷‍♂️

PR #291, opened 2018-07-25
Navigating back to 📅 after having already saved dates

nmakuch:
# This PR consists of:
## Making date buckets appear on top of the calendar when navigating back to the page after having already saved some values:

| Before | After |
|--------|-------|
|    topcal-buckers-before    |   topcal-buckers-after    |

## Removed some default button padding that was being applied to the remove date icon:

| Before | After |
|--------|-------|
|  screen shot 2018-07-25 at 11 49 32|  screen shot 2018-07-25 at 11 49 05  |


web/src/components/Calendar.js
```diff
@@ -484,6 +527,7 @@ class Calendar extends Component {
     this.state = {
       errorMessage: null,
     }
+    this.threeDatesArePicked = this.props.input.value.length === 3
```
pcraig3: 😉 this is only working incidentally. 
`input.value` is an empty string before any dates have been selected.

A more robust check would be more like:
- does value exist
- is it an array
- is it length === 3

web/src/components/Calendar.js
```diff
@@ -419,6 +423,45 @@ const triangle = css`
   `)};
 `
 
+const reviewContainer = css`
```
pcraig3: `reviewContainer` only has one new rule that `calendarContainer` doesn't have: https://www.diffchecker.com/p5rq8p67

To save duplication, you could do somethign like:
```
const calendarContainer = css`
... styles ...
```

const calendarContainerBottom = css`
 ${calendarContainer};
 flex-direction: forwards; 
`


const calendarContainerTop = css`
 ${calendarContainer};
 flex-direction: reverse; 
`
```

👆 that's psuedo code, but you get the idea.

PR #292, opened 2018-07-25
Add <titles> to pages

pcraig3:
We've had a couple of accessibility audits now that have pointed out that our `<title>`s don't change as we go between the different pages of our app. 

No longer!

This PR:
- adds a `Title` component
- adds some tests for it
- adds `Title` to all of our pages
- removes the Helmet stuff from the `LanguageSwitcher`
- works in english and french


pcraig3:
🚢

PR #294, opened 2018-07-26
Set up the sentry logging for source maps

Amirroracle:
In this PR:
-Proper release definitions mapped by environment variables added in ```Document.js``` and ```Server.js```.
-Add sentry cli command that will enable artifact deployment to sentry.
-Cli command still needs to be added to the build deployment scripts.

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -12,12 +12,14 @@
     "extract": "NODE_ENV=development lingui extract",
     "compile": "NODE_ENV=development lingui compile",
     "add-locale": "lingui add-locale",
-    "stats": "BUNDLE_CHECK=true yarn build"
+    "stats": "BUNDLE_CHECK=true yarn build",
+    "push-source-map": "sentry-cli releases  -o canadian-digital-service -p ircc-rescheduler files  RAZZLE_STAGE  upload-sourcemaps ../build/"
   },
   "dependencies": {
     "@cdssnc/gcui": "^0.0.30",
     "@jaredpalmer/after": "^1.3.1",
     "aws-sdk": "^2.279.1",
+    "@sentry/cli": "^1.34.0",
     "body-parser": "^1.18.3",
     "cookie-encrypter": "^1.0.1",
     "cookie-parser": "^1.4.3",
--- a/web/razzle.config.js
+++ b/web/razzle.config.js
@@ -25,19 +25,19 @@ module.exports = {
         }),
       )
     }
-    if (process.env.RAZZLE_SENTRY_API) {
+    /* if (process.env.RAZZLE_SENTRY_API) {
       const SentryPlugin = require('@sentry/webpack-plugin')
       config.plugins.push(
         new SentryPlugin({
           organization: 'canadian-digital-service',
           project: 'ircc-rescheduler',
-          release: 'a2315885b9c3429a918336c1324afa4a',
+          release: 'process.env.RAZZLE_SENTRY_STAGE',
           include: '.',
           ignore: ['node_modules', 'webpack.config.js'],
           apiKey: process.env.RAZZLE_SENTRY_API,
         }),
       )
-    }
+    }*/
     return config
   },
 }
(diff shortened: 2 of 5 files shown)


web/razzle.config.js
```diff
@@ -25,19 +25,19 @@ module.exports = {
...
     }
-    if (process.env.RAZZLE_SENTRY_API) {
+    /* if (process.env.RAZZLE_SENTRY_API) {
       const SentryPlugin = require('@sentry/webpack-plugin')
       config.plugins.push(
         new SentryPlugin({
           organization: 'canadian-digital-service',
           project: 'ircc-rescheduler',
-          release: 'a2315885b9c3429a918336c1324afa4a',
+          release: 'process.env.RAZZLE_SENTRY_STAGE',
```
pcraig3: I think this should be `RAZZLE_STAGE`

web/razzle.config.js
```diff
@@ -25,19 +25,19 @@ module.exports = {
         }),
       )
     }
-    if (process.env.RAZZLE_SENTRY_API) {
+    /* if (process.env.RAZZLE_SENTRY_API) {
```
pcraig3: Wait, is this whole code block commented out?

web/package.json
```diff
@@ -12,12 +12,14 @@
     "extract": "NODE_ENV=development lingui extract",
     "compile": "NODE_ENV=development lingui compile",
     "add-locale": "lingui add-locale",
-    "stats": "BUNDLE_CHECK=true yarn build"
+    "stats": "BUNDLE_CHECK=true yarn build",
+    "push-source-map": "sentry-cli releases  -o canadian-digital-service -p ircc-rescheduler files  RAZZLE_STAGE  upload-sourcemaps ../build/"
```
pcraig3: Does this run locally, or we need to make sure we have our auth token set?

Amirroracle: we need to have our auth token set, but anyone can run this locally if they have it.

web/razzle.config.js
Amirroracle: Keeping it commented out in case we need to use it again, I can remove it if it's ugly

web/package.json
timarney: Should we update the Readme with instructions about setting the auth token + the new env var?

web/razzle.config.js
timarney: I think we should remove it.

timarney:
Can we assume as is the case with the server (~/server.js.map) for stage / production get a proper single client source map?

Just check on Sentry and all the got reloading map files are showing up.
```
~/0.09ae37e5862226cfddf1.hot-update.js
```

web/package.json
Amirroracle: I believe @pcraig3  already did that :)

web/razzle.config.js
Amirroracle: 👍

web/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "ircc-rescheduler",
-  "version": "1.0.1",
+  "version": "1.5.0",
```
pcraig3: can we do `1.1.0` or `1.0.2`?

Amirroracle: Yeah this was an attempt to fix the circle ci stuff, I'll revert.

web/src/Document.js
```diff
@@ -69,7 +69,9 @@ class Document extends React.Component {
           {process.env.NODE_ENV === 'production' && (
             <script
               dangerouslySetInnerHTML={{
-                __html: `Raven.config('https://a2315885b9c3429a918336c1324afa4a@sentry.io/1241616').install()`,
+                __html: `Raven.config('https://a2315885b9c3429a918336c1324afa4a@sentry.io/1241616', {release: ${
+                  process.env.RAZZLE_STAGE
```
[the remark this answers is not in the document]
Amirroracle: good catch! Yeah I'll definitely add null checks here and in server.js. I don't want to default the name to an existing release value so that way if the env var is null, it becomes apparent :)

PR #295, opened 2018-07-26
Sort out site headers

pcraig3:
Our accessibility audits are mad at us for having something that looks visually like a page title but isn't one.

After talking with @zakkain, we decided to remove the "Request a new citizenship appointment" title from all pages except the first page, and on the first page we're making it an H1 because it's useful information you actually want to read.

That means that most pages now will just have the "skinny" blue banner with the beta tag in it.

# This PR

- removes the "Request a new citizenship appointment" title from most pages
- updates a couple page titles based on Zak's slack update yesterday (it's in the commit message)
  - changes registration page message
  - changes the header on the calendar page
- both in english and french

(the diffs are a bit hard to read after doing the extract + compile, so if you check out the first two commits directly it might be easier)

# Screens
| before | after |
|--------|-------|
| screen shot 2018-07-26 at 5 29 02 pm  |  screen shot 2018-07-26 at 5 28 54 pm |
|  screen shot 2018-07-26 at 5 29 58 pm |  screen shot 2018-07-26 at 5 29 55 pm  |
| screen shot 2018-07-26 at 5 30 05 pm  | screen shot 2018-07-26 at 5 30 07 pm   |


timarney:
This solves the problem nicely.

On the calendar page:

Should we be looking at these as well (maybe a separate PR)?  H2's either hidden or empty.

|   Selected Day List	|   Error Text	
|---	|---	
|   screen shot 2018-07-27 at 7 27 23 am	|   screen shot 2018-07-27 at 7 28 44 am

PR #296, opened 2018-07-30
Cypress End to end

timarney:
Initial End to End test setup.

This adds a single end to end tests for the app and also create a "next" version of the circleci config file.

The "next" config will eventually replace the existing config.  Some further discussion needs to be had on the config setup / environment variables prior to switching to the "next" version. 

Getting this in now will allow us to start using Cypress for testing locally.

New scripts commands have been added ci:dev and ci:prod to start and run the cypress tests.


web/cypress/integration/main/cal.spec.js
```diff
@@ -0,0 +1,59 @@
+context('Calendar Date selections', () => {
+  beforeEach(() => {
+    cy.visit('/')
+  })
+
+  it('should be able to select and unselect dates', () => {
```
pcraig3: This is super good. 

Really easy to see what these tests are doing.

web/src/pages/CalendarPage.js
```diff
@@ -237,7 +237,7 @@ class CalendarPage extends Component {
             }
 
             return (
-              <form onSubmit={handleSubmit} className={fullWidth}>
+              <form id="cal-form" onSubmit={handleSubmit} className={fullWidth}>
```
pcraig3: Small thing: can we call this `id="calendar-form"` to match `id="calendar-header"`?

web/cypress/integration/main/form.spec.js
```diff
@@ -0,0 +1,54 @@
...
+  beforeEach(() => {
+    cy.visit('/')
+  })
+
+  it('should be able to fill in a profile and reach the confirmation page', () => {
+    cy.get('main a span').should('have.text', 'Start now')
+    cy.get('main a').click({ force: true })
+
+    cy.fixture('user').then(data => {
+      cy.get('#fullName').type(data.fullName, { force: true })
```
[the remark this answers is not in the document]
timarney: Force true is being used to get around the "span" button issue where there is an item overlapping the item you want to click.

PR #297, opened 2018-07-30
Add automated a11y tests. 

timarney:
Script will run through a given set of urls and scan for issues.

New command added to scripts a11y:test will kick off the scan.

Issues will be output to the console for user review.

This PR does a basic scan of the pages / urls.  Moving forward the tests can fill in data and check for things like form errors. This will be done using "actions" a sample of which has been included.

i.e.

const registerActions = [
  'set field #fullName to John Doe',
  'click element main button',
]

screen shot 2018-07-30 at 8 11 40 am

### DIFF
--- a/web/a11y.test.js
+++ b/web/a11y.test.js
@@ -21,6 +21,11 @@ const visit = [
   { url: '/register', registerActions },
   { url: '/calendar' },
   { url: '/review' },
+  { url: '/confirmation' },
+  { url: '/confirmation/error' },
+  { url: '/cancel' },
+  { url: '/error' },
+  { url: '/404' },
 ]
 
 async function run() {
--- a/web/a11y.test.js
+++ b/web/a11y.test.js
@@ -1,0 +1,48 @@
+const pa11y = require('pa11y')
+const baseUrl = 'http://localhost:3004'
+const options = {
+  log: {
+    debug: console.log,
+    error: console.error,
+    info: console.log,
+  },
+}
+const registerActions = [
+  'set field #fullName to John Doe',
+  'click element main button',
+]
+
+/*--------------------------------------------*
+ * List of urls we want to visit
+ *--------------------------------------------*/
+
+const visit = [
+  { url: '/' },
+  { url: '/register', registerActions },
+  { url: '/calendar' },
+  { url: '/review' },
+]
+
+async function run() {
+  try {
+    const results = await Promise.all(
+      visit.map(page => {
+        return pa11y(baseUrl + page['url'], {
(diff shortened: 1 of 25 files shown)


web/a11y.test.js
```diff
@@ -0,0 +1,53 @@
...
+  { url: '/404' },
+]
+
+async function run() {
+  try {
+    const results = await Promise.all(
+      visit.map(page => {
+        return pa11y(baseUrl + page['url'], {
+          ...options,
+          ...{ actions: page['actions'] ? page['actions'] : [] },
```
pcraig3: Is `page['actions']` going to return anything here? 

It looks like `page` will be `{ url: '/register', registerActions }`. 
Should it be `{ url: '/register', actions: registerActions }` instead?

web/package.json
```diff
@@ -76,8 +79,10 @@
     "jest": "^23.4.1",
     "jest-cli": "^23.4.1",
     "lingui-cli": "^1.4.4",
-    "prettier": "^1.13.7",
+    "pa11y": "^5.0.4",
+    "prettier": "^1.14.0",
     "raf": "^3.4.0",
+    "start-server-and-test": "^1.7.0",
```
pcraig3: This is pretty good, eh? Saves a lot of setup work.

web/package.json
```diff
@@ -8,11 +8,14 @@
...
     "lint": "node_modules/eslint/bin/eslint.js src test",
     "start": "NODE_ENV=production node --icu-data-dir=./node_modules/full-icu build/server.js",
+    "build:start": "yarn build && yarn start",
     "test": "node --icu-data-dir=./node_modules/full-icu node_modules/jest/bin/jest.js",
     "extract": "NODE_ENV=development lingui extract",
     "compile": "NODE_ENV=development lingui compile",
     "add-locale": "lingui add-locale",
     "stats": "BUNDLE_CHECK=true yarn build",
+    "a11y:test" : "start-server-and-test build:start http://localhost:3004 a11y:run",
+    "a11y:run" : "node a11y.test.js",
```
pcraig3: The only error I get when I run this directly (and I'm not running the app locally) is `net::ERR_CONNECTION_REFUSED at http://localhost:3004/404`.

Kind of weird you don't see it for the other pages, but 🤷‍♂️

web/a11y.test.js
pcraig3: As a followup, I made this change locally and I don't see any new errors (yay!) but there is a new console message.

- `✔︎ action complete`

So just missing a key, it looks like.

PR #285, opened 2018-07-23
Translating ABBR tag titles

nmakuch:
**A lot of useful information was pulled from an** [ODS article](https://www.ontario.ca/document/ontario-design-guide/abbreviations)
# This PR addresses the following [Trello Card](https://trello.com/c/R72YUPDp): 

## Adding explicit underline to abbreviations:
* Underline was breaking into a weird jumble in the previous version

| Before | After |
|--------|-------|
|   screen shot 2018-07-23 at 16 28 33     |     screen shot 2018-07-23 at 16 29 02   |

## Add explicit mouse cursor CSS rule to abbreviations:

| Before | After |
|--------|-------|
|    abbr-cursor-before | abbr-cursor-after    |

## Abbr's auto-expand when in mobile mode:

| Before | After |
|--------|-------|
|    abbr-expand-before    |  abbr-expand-after |

## Abbreviation text switches back and forth in english and french:

| Before | After |
|--------|-------|
|    abbr-language-before    |   abbr-language-after    |

PR #298, opened 2018-07-30
Adds space between "your local" and IRCC on the confirmation page.

timarney:
| Before | After |
|--------|-------|
| screen shot 2018-07-30 at 8 30 05 am |  screen shot 2018-07-30 at 8 31 23 am

  |

### DIFF
--- a/web/src/pages/ConfirmationPage.js
+++ b/web/src/pages/ConfirmationPage.js
@@ -106,7 +106,7 @@ class ConfirmationPage extends React.Component {
           </H2>
           <p>
             <Trans>By</Trans> {respondBy}
-            , <Trans>your local</Trans>
+            , <Trans>your local</Trans>{' '}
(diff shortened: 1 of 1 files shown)


pcraig3:
I [alluded to this ~bug~ not actually a bug](https://github.com/cds-snc/ircc-rescheduler/pull/285#pullrequestreview-140035619) when the abbreviation PR was merged. 

Adding an extra space throws off our french translation, ie we end up with `d‘ IRCC` on the page, which isn't right. So [we added an extra space to the english messages.json file](https://github.com/cds-snc/ircc-rescheduler/blob/master/web/locale/en/messages.json#L654). For some reason, it doesn't show up when you run `yarn dev` (the extra space is missing), but it _does_ show up when you run `yarn start`. 

So the app when it's deployed always looks good. When developing locally, we just have to grit our teeth and ignore it. :(

PR #304, opened 2018-07-30
New circleci config runs cypress tests

pcraig3:
New circleci config will run our cypress tests (and possibly our a11y tests, except currently they will fail us so we should fix them first).


pcraig3:
everything has passed 🏁 🏁 🏁 

image

PR #307, opened 2018-07-31
Update review days to show sorted dates selected

timarney:
This fixes the sort order of the dates on the Review Page. 

Previously this page would display dates in the order that the user selected the dates.

**Note:**
Ultimately it would be nice to have a single function to handle this for all the possible output (Calendar, Review Page + Email).  Maybe updating sortSelectedDays to check if the items in the array are dates already.  Would need to determine what each page / component is looking for.


pcraig3:
👍

PR #310, opened 2018-07-31
Accidentally added a space to the build path

pcraig3:
🌔 🌟 🌟 🚀 👨‍🚀

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -20,10 +20,10 @@
     "cypress:run": "cypress run --record",
     "ci:dev": "start-server-and-test dev http://localhost:3004 cypress:open",
     "ci:prod": "start-server-and-test build:start http://localhost:3004 cypress:run",
-    "sentry_auth": "DATETIME=`date +%Y%m%d%H%M%S` && . $(readlink .env.production) && SENTRY_AUTH_TOKEN=$SENTRY_AUTH_TOKEN ",
+    "sentry_auth": "COMMIT_HASH=`git rev-parse --short HEAD` && . $(readlink .env.production) && SENTRY_AUTH_TOKEN=$SENTRY_AUTH_TOKEN ",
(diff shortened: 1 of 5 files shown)


web/package.json
```diff
@@ -12,17 +12,18 @@
...
     "cypress:open": "cypress open",
     "cypress:run": "cypress run --record",
     "ci:dev": "start-server-and-test dev http://localhost:3004 cypress:open",
     "ci:prod": "start-server-and-test build:start http://localhost:3004 cypress:run",
-    "sentry-map": "sentry-cli releases -o canadian-digital-service -p ircc-rescheduler files  $RAZZLE_STAGE  upload-sourcemaps ./build/ --rewrite",
-    "sm:stg": "source .env.production && SENTRY_AUTH_TOKEN=$SENTRY_AUTH_TOKEN RAZZLE_STAGE=staging yarn sentry-map",
-    "sm:prod": "source .env.production && SENTRY_AUTH_TOKEN=$SENTRY_AUTH_TOKEN RAZZLE_STAGE=production yarn sentry-map"
+    "sentry_auth": "DATETIME=`date +%Y%m%d%H%M%S` && . $(readlink .env.production) && SENTRY_AUTH_TOKEN=$SENTRY_AUTH_TOKEN ",
+    "sentry_map": "sentry-cli releases -o canadian-digital-service -p ircc-rescheduler files  $RAZZLE_STAGE  upload-sourcemaps ./build/ --rewrite",
+    "sm:stg": "yarn sentry_auth RAZZLE_STAGE=staging-$DATETIME yarn sentry_map",
```
Amirroracle: Only problem with ```RAZZLE_STAGE=staging-$DATETIME1``` is that the date-time variable is not going to match with what is in ```server.js``` and ```document.js```. The release strings have to be the same, so what will happen is for instance you will run a ```yarn build```, deploy the app etc... Then you will run ```sm:stg``` and that will upload the source map artifacts to a different release since it will have a date-time string.

timarney: Wonder if we could pull this off - git rev-parse --short HEAD

PR #311, opened 2018-07-31
test auto-run a11y tests

timarney:


.circleci/config.yml
```diff
@@ -35,6 +35,7 @@ jobs:
 
       # run tests!
       - run: yarn build
+      - run: yarn a11y:test
```
pcraig3: Do we want to always run this first? For me, it's okay either way -- just wondering what you think.

timarney: I think that's fine.

On another note:
We now have multiple scripts that build and run.  Should be able to build once and removing building from the other scripts but will save that for another separate PR.

pcraig3: Yeah, that's a good point. We could speed up our tests by not building every time.

PR #315, opened 2018-08-01
Listen for "accept-language" browser header

pcraig3:
If someone explicitly sets a default language, or passes in a "language" query parameter, we save that to a cookie.

However, when they come to our site for the first time:
- they won't have any cookies
- they probably aren't passing in a query param
- they may have a language setting in their browser

So that's the case we want to check.

If there are no cookies and their "Accept-Language" header (comes through as lowercase in express) starts with either "en", or "fr", then we'll pick a default language for them.

Note that it is possible someone will have a header like:
- `"es,en-US,en"`
- `"zh, fr-CA, en"`

In both of those cases, the string "fr" or "en" exists in their accept-language header, but it's not being picked up.

Since I doubt that that is happening to us, I am opting for simpler logic and not handling that case.

Also wrote a bunch of tests to check for various possible header values.

### DIFF
--- a/web/src/__tests__/WithProvider.test.js
+++ b/web/src/__tests__/WithProvider.test.js
@@ -227,6 +227,14 @@ describe('WithProvider', () => {
         expect(result).toEqual({ field: 'value' })
       })
 
+      it('returns trimmed string values when validation passes trimmed value', () => {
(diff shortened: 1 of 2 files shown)


pcraig3:
Removed the last commit as the package has already been updated.

PR #316, opened 2018-08-01
Track language toggle

timarney:
Sends Google Analytics event to track when users click the language toggle link.

Adds a label parameter to the logEvent function so we can track value

i.e. logEvent('Navigation', 'Toggle Language', 'FR')

Also added a few checks for window as Razzle was complaining.  Likely a timing issue.

screen shot 2018-08-01 at 2 58 25 pm

### DIFF
--- a/web/src/components/LanguageSwitcher.js
+++ b/web/src/components/LanguageSwitcher.js
@@ -12,6 +12,8 @@ import {
   focusRing,
 } from '../styles'
 
+import { logEvent } from '../utils/analytics'
+
 const button = css`
   ${focusRing};
   display: inline-block;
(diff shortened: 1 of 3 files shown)


pcraig3:
Just tried this out locally, and it seems to be sending 7 events each time the language is switched.

screen shot 2018-08-02 at 9 35 53 am

It says I have 28 events after clicking the link 1 time. As you can see in the bar graph, each event is showing up as "7".

I checked the events that came through yesterday and they are also both multiples of 7.

screen shot 2018-08-02 at 9 35 13 am

If we can't figure out why this is happening, we could just caveat the event totals by saying they should be divided by 7, but it's probably worth looking into before this is merged.

PR #321, opened 2018-08-02
Google Analytics tracking for form errors

timarney:
Adds event tracking for the Calendar and Registation pages

1. Will log Calendar Submit event for errors. The event will be submitted with a label containing the number of days selected by the user.

2. Will send a Registration Submit Event with a label matching the field errors i.e.
Error(s): fullName, Email..., paperFileNumber, Reason

### DIFF
--- a/web/src/pages/CalendarPage.js
+++ b/web/src/pages/CalendarPage.js
@@ -37,6 +37,7 @@ import CalendarNoJS from '../components/CalendarNoJS'
 import CancelButton from '../components/CancelButton'
 import { Checkbox } from '../components/forms/MultipleChoice'
 import { getEndMonthName, getStartMonthName } from '../utils/calendarDates'
+import { logEvent } from '../utils/analytics'
 
 const DAY_LIMIT = 3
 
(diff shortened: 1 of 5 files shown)


web/src/pages/CalendarPage.js
```diff
@@ -161,6 +162,8 @@ class CalendarPage extends Component {
   async onSubmit(values, event) {
     const submitErrors = this.validate(values)
 
+    logEvent('Calendar', 'Submit', `Error: ${values.selectedDays.length} Day(s) selected`)
```
pcraig3: can we move these log events down into the `if` check just below? 
After the `if (Object.keys(submitErrors).length) {` check, we know we have errors.

web/src/pages/RegistrationPage.js
```diff
@@ -140,6 +141,8 @@ class RegistrationPage extends React.Component {
   async onSubmit(values, event) {
     const submitErrors = this.validate(values, true)
 
+    trackRegistrationErrors(submitErrors)
```
pcraig3: Same here, these could be moved into the `if` check and that way only get triggered when the error object is not empty.

web/src/pages/CalendarPage.js
pcraig3: There's also, in the `Calendar.js` file, the "too many days" error. Worth tracking that one as well do you think?

PR #322, opened 2018-08-02
Fixed Wordbreak on confirmation mobile screens

### LINKED ISSUE #319: Email looks like sucks
The Vancouver IRCC email address is super long: IRCC.DNCitVANNotification-NotificationVANCitRN.IRCC@cic.gc.ca

We put in some [CSS word-break stuff](https://css-tricks.com/snippets/css/prevent-long-urls-from-breaking-out-of-container/) so that the email will fit on the screen. So now the email breaks across lines, but we don't really decide how that's happening.

| before | after |
|--------|-------|
| image from ios  |  image from ios 1  |

Didn't really make it better.

**Proposed solution:**

Let's:
- take out the word-break CSS
- make the link a block-level element
- use [`<wbr>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/wbr) elements so that the email breaks where we want it (on mobile).

Like this:
```html
<a href="mailto:IRCC.DNCitVANNotification-NotificationVANCitRN.IRCC@cic.gc.ca">
  IRCC.DNCitVANNotification-<wbr />NotificationVANCitRN.IRCC<wbr />@cic.gc.ca
</a>
```

Should look like this:
screen shot 2018-08-01 at 10 10 59 pm

👍

nmakuch:
# This PR consists of:

## Fixing the horrendous wordbreak on the confirmation page:

| Before | After |
|--------|-------|
|   screen shot 2018-08-02 at 14 01 08     |   screen shot 2018-08-02 at 14 01 56    |

(Merging this PR closes #319 )

### DIFF
--- a/web/src/components/Contact.js
+++ b/web/src/components/Contact.js
@@ -43,8 +43,8 @@ const contact = css`
   }
 
   ${mediaQuery.xs(css`
-    a[id='email'],
-    a[id='telephone'] {
+    a#email,
+    a#telephone {
       display: block;
     }
     wbr {
--- a/web/src/components/Contact.js
+++ b/web/src/components/Contact.js
@@ -24,7 +24,7 @@ const telStyles = css`
 const TelLink = ({ tel }) => (
   <span className={telStyles}>
     <span>{tel}</span>
-    <a name="telephone" href={`tel:+${tel}`} rel="nofollow">
+    <a id="telephone" href={`tel:+${tel}`} rel="nofollow">
       {tel}
     </a>
   </span>
@@ -42,9 +42,9 @@ const contact = css`
     display: none;
   }
 
-  ${mediaQuery.sm(css`
-    a[name='email'],
-    a[name='telephone'] {
+  ${mediaQuery.xs(css`
+    a[id='email'],
+    a[id='telephone'] {
       display: block;
     }
     wbr {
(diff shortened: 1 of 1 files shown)


web/src/components/Contact.js
```diff
@@ -24,7 +24,7 @@ const telStyles = css`
 const TelLink = ({ tel }) => (
   <span className={telStyles}>
     <span>{tel}</span>
-    <a href={`tel:+${tel}`} rel="nofollow">
+    <a name="telephone" href={`tel:+${tel}`} rel="nofollow">
```
pcraig3: uh oh, looks like `name` is deprecated in HTML5.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#Obsolete_attributes

Alternatively, you could just use `id="telephone"` or `class="telephone"`

web/src/components/Contact.js
```diff
@@ -35,23 +35,22 @@ TelLink.propTypes = {
...
+  wbr {
+    display: none;
+  }
+
+  ${mediaQuery.sm(css`
+    a[name='email'],
+    a[name='telephone'] {
+      display: block;
+    }
+    wbr {
```
pcraig3: Can we do the `wbr` rule only on `xs`?

web/src/components/Contact.js
```diff
@@ -35,23 +35,22 @@ TelLink.propTypes = {
...
   p:first-of-type {
     margin-bottom: ${theme.spacing.sm};
   }
+  wbr {
+    display: none;
+  }
+
+  ${mediaQuery.xs(css`
+    a[id='email'],
+    a[id='telephone'] {
```
pcraig3: now that you're using ids, you can be like `a#telephone`, but it's fine either way, so just merging as-is is fine with me.

PR #324, opened 2018-08-03
Calendar Error Regression

timarney:
Fixes issue where Calendar errors for too many days selected ends up behind by one click.

The previous fix for this used forced render which fixed the error issue but, created an issue where users could not unselect dates after they we're saved to the cookie.

In this version we're passing the selected stated back to the force render and than updating state (using either the saved or state for the Calendar initial values).

This allows the Calendar to do a full re-render while maintaing the correct dates.

### DIFF
--- a/web/src/components/Calendar.js
+++ b/web/src/components/Calendar.js
@@ -538,7 +538,6 @@ class Calendar extends Component {
 
     // !selected means that this current day is not marked as 'selected' on the calendar
     if (!selected) {
-      console.log('force render', selectedDays.length)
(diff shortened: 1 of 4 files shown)


web/src/components/Calendar.js
```diff
@@ -571,7 +571,8 @@ class Calendar extends Component {
       errorMessage: null,
     })
 
-    //this.props.forceRender()
+    // force a render to keep calendar errors in sync
+    this.props.forceRender(selectedDays)
```
pcraig3: This is really clean. A+

PR #327, opened 2018-08-03
Add heroku config files

pcraig3:
- Adds all the env vars we need to run the app
- Adds a buildpack that should let us deploy from a subdirectory

Note: added two files since I'm not sure which one heroku will pick up.


pcraig3:
Looks like this needs to be merged before I can tinker with PR deploys. Will remove the unneeded file once I can get something working.

PR #336, opened 2018-08-08
Url params matching issue

timarney:


web/src/utils/url.js
```diff
@@ -0,0 +1,25 @@
...
+        .split('&')
+        .reduce((params, param) => {
+          let [key, value] = param.split('=')
+          params[key] = value
+            ? decodeURIComponent(value.replace(/\+/g, ' '))
+            : ''
+          return params
+        }, {})
+    : {}
+}
```
pcraig3: Yikes, I hope you found this somewhere rather than having to write one yourself.

PR #341, opened 2018-08-08
Version B - Usability Testing

nmakuch:


web/src/components/MobileCancel.js
```diff
@@ -0,0 +1,30 @@
...
+            d="M11.93 10.82l3.56-3.56a.79.79 0 0 0 0-1.11.8.8 0 0 0-1.12 0L10.82 9.7 7.26 6.15a.8.8 0 0 0-1.12 0 .79.79 0 0 0 0 1.11l3.55 3.55-3.55 3.55a.79.79 0 0 0 .55 1.34.8.8 0 0 0 .56-.23l3.55-3.54 3.55 3.55a.79.79 0 0 0 .56.23.77.77 0 0 0 .54-1.33z"
+          />
+        </g>
+      </g>
+    </svg>
+  )
+}
+
+MobileCancel.propTypes = {
+  circleColour: PropTypes.string, //left or right
```
pcraig3: Don't need this comment here.

web/src/components/Calendar.js
```diff
@@ -397,6 +398,10 @@ const removeDate = css`
   width: 1rem;
   height: 1rem;
 
+  img circle {
+    fill: red;
+  }
+
```
pcraig3: Looks like this CSS isn't being used anymore. Is it?

web/src/pages/ConfirmationPage.js
```diff
@@ -108,10 +96,9 @@ class ConfirmationPage extends React.Component {
             <Trans>What happens next?</Trans>
           </H2>
           <p>
-            <Trans>By</Trans> {respondBy},{' '}
             <Trans>
-              we’ll send you a new appointment. You will always be contacted at
-              least 3 weeks before your appointment.
+              IRCC will send you a new appointment. You will always be contacted
+              at least 3 weeks before your appointment.
```
pcraig3: If we're changing this sentence, should we change it in the email as well?

web/src/pages/ConfirmationPage.js
```diff
@@ -108,10 +96,9 @@ class ConfirmationPage extends React.Component {
             <Trans>What happens next?</Trans>
           </H2>
           <p>
-            <Trans>By</Trans> {respondBy},{' '}
             <Trans>
-              we’ll send you a new appointment. You will always be contacted at
-              least 3 weeks before your appointment.
+              IRCC will send you a new appointment. You will always be contacted
```
pcraig3: Also, we can bring back our IRCCAbbr now that we're referring to IRCC again.

web/src/components/Calendar.js
nmakuch: Woops, I had removed that but then went back during some last minute changes before usability testing and it must have gotten in somehow :s nice catch

web/src/pages/ConfirmationPage.js
nmakuch: true! good point

nmakuch: tried to implement it to the email as well but after reading up on the subject I don't think most email applications support pseudo-selectors.

PR #344, opened 2018-08-09
Add tests that include tripping validation errors

pcraig3:
We recently had a problem where validation errors would cause values not be saved.

We've since released a fix for this, and these tests are confirming the correct behaviour.

It's not doing a lot of code reuse, so I can do some refactoring if needed.

## gif

test-1

### DIFF
--- a/web/cypress/fixtures/user.json
+++ b/web/cypress/fixtures/user.json
@@ -3,5 +3,6 @@
 "fullName": "Leanne Graham",
 "email": "cdc@example.com",
 "paperFileNumber": "123456",
+"reason": "workOrSchool",
 "explanation": "On business trip!"
 }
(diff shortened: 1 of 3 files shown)


pcraig3:
Hey @timarney, take a look at what the new query parameter tests are doing.

For the calendar page in particular, it's not really testing what a user would actually do (we're not getting the checkbox calendar interface, for example), but it _is_ testing what it's possible to do with our app.

If you think it's too weird, I can just go back to clicking the calendar as per usual once we're past the registration page.

How do you feel about it?

pcraig3:
Talked about this on slack, but this test isn't actually triggering the "too many days selected" error.

It triggers the "not enough days" error, and then it submits the form, so it only needs 3 days to be selectable.

So I'm going to merge it as-is.

PR #346, opened 2018-08-09
French end-to-end test

timarney:


web/cypress/integration/main/full-run-fr.js
```diff
@@ -0,0 +1,49 @@
+export const fullRun = cy => {
```
pcraig3: It looks like the only difference between this file and `full-run.js` is the `'Commencer'` string and the `'user-fr'` string. Could we merge these together somehow?

Edit: ah, just noticed the language toggle as well. Could hide that behind an `if` (ie, if "commencer" vs "start now"), but up to you.

PR #347, opened 2018-08-10
gitHash returns hash as variable rather than logging it

pcraig3:
## Return hash value instead of logging it

By setting the returned value to `process.stdout`, it seems like the result of the execSync call was being logged to the screen while the value returned to the app was `000000`.

[The docs say that the default value is 'pipe'](https://nodejs.org/api/child_process.html#child_process_child_process_execsync_command_options), so I've reset the first two std values to 'pipe'

Note: I tried to write tests for this, but ran into two problems
1. I can't predict whether this will be run from a git repo or not, so I don't know what value to expect
2. I couldn't figure out how to get the logged hash from the original function while still in a test

I think it's okay as-is, but open to suggestions.


pcraig3:
A+

PR #351, opened 2018-08-10
French text updates

timarney:


web/locale/fr/messages.json
```diff
@@ -1,10 +1,64 @@
...
+    "translation": "Les lundis",
+    "origin": [
+      [
+        "src/utils/calendarDates.js",
+        221
+      ]
+    ]
+  },
+  "Tuesdays": {
+    "translation": "Les mardis",
```
pcraig3: I think these should be lower case in french.

PR #352, opened 2018-08-10
Remove cookie encryption

pcraig3:
#  Remove cookie encryption
We don't need it.

- Cookies are only served to our rescheduler domain (means that other
  sites can't look at them).
- Our rescheduler domain only sends cookies over HTTPS, so the site
  traffic is encrypted end-to-end

There shouldn't be a way to actually get the content of the cookies
unless you're the person who entered the values in the first place.

### DIFF
--- a/web/README.md
+++ b/web/README.md
@@ -63,8 +63,6 @@ RAZZLE_COOKIE_HTTP=true
 
 These options are set when running in production mode (`yarn start`), whether locally or on the server.
 
-- `RAZZLE_COOKIE_SECRET`: A secret used to encrypt user data. Must be 32 characters long. If left unset, a default secret will be used (ie, for local development).
-
 - `RAZZLE_GA_ID`: Our Google Analytics ID code. If left unset, the app won’t send pageviews to Google. Not generally needed during local development but can be turned on to test functionality.
 
 - `SENTRY_AUTH_TOKEN`: In order to upload our sourcemaps to sentry, we have to set an auth token. We upload source maps as part of deploys so that we can trace errors back to specific versions. This is not used once the app is running, so it doesn't use the `RAZZLE_` prefix.
@@ -72,7 +70,6 @@ These options are set when running in production mode (`yarn start`), whether lo
 ##### sample `web/.env.production` file
 
 ```
-RAZZLE_COOKIE_SECRET='DONE_KEPT_IT_REAL_FROM_THE_JUMP_'
 RAZZLE_GA_ID='UA-111111111-1'
 SENTRY_AUTH_TOKEN='notARealAuthToken'
 ```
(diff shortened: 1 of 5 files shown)


pcraig3:
Here's my writeup of this:

### a traditional app 

In a traditional app (client->server->client->server->client, etc.), you could keep all user data in a session. You would save the session ID in a cookie, and but then all the user data would go into a session, where the user wouldn't be able to touch it. ie, like that 👇 

server-response 1

### our app 

Our Single Page App, calls the server on the first request, but after that it does pretty much all the work in the client (client->server->client->client->client, etc.). This means we can't store anything on the server because those changes won't be propagated as we continue to make changes in the client. 

Our client state remains consistent, but as soon as the JavaScript is interrupted (ie, doing a refresh or something), we lose everything. All your data and language preference and everything gets lost.

To get around this, we're setting cookies. We can access them on the client so every time we update the internal state, we set this in a cookie. But we can also read and set cookies on the server, so if we do happen to do a full-page reload halfway through, all the cookie data is accessed by the server and used to rehydrate the app.

client-response-2

### mitigations

- the cookies are only served to our domain, so other servers can't see it.
- the cookies are only served over HTTPS, so all the in between traffic is encrypted

This means that the only person with access to the cookie should be the client themselves.

### unencrypting cookies

We've been using encrypted cookies so that the cookie values remain inscrutable to users. They can use our app to save values, but since they can't mess with the cookie values themselves.

Encrypted cookie 👇 

screen shot 2018-08-13 at 3 00 53 pm

Not-encrypted cookies are in plaintext 👇 

screen shot 2018-08-13 at 3 00 18 pm

Once this PR goes in, we'll have plaintext, unsigned cookies. This means that malicious users could tamper with them before sending them back. 

### options

1. Look for some other encryption package if we want to keep cookies super locked down. I've seen a couple implementations of encrypted cookies (by [mozilla](https://github.com/mozilla/node-client-sessions), or [this other one](https://github.com/hex7c0/cookie-encryption)), but most are expecting to be run on the `req`/`res` objects that we only have on the server. Also, if we do find something that will work in IE11, we are likely stuck with a ballooning bundle size.

 2. Use signed cookies instead of unsigned cookies. This means that they will be tamper proof, although they will still be visible. I tried setting this up initially and was foiled by trying to read signed cookies on the client, but there must be a way.

3. Keep it as-is, and accept the risk. Tampering with a cookie only means you can get around our validation, and if you get around that you are only hurting your chances of getting the email or send or having your appointment rebooked. All the information is the user's anyway, and it expires in an hour. This can't be exploited to get other users' data or send a bunch of weird emails or something, so there's no real incentive to game this.

pcraig3:
Going with (3) ☝️.

PR #357, opened 2018-08-13
Fix empty error message when submitting no days

nmakuch:
## This PR consists of:

- We had fixed this earlier, but now it's showing a blank error message again, so added a test case that submits with no days selected and looks for the error message.

## Error Message was showing up blank when submitting calendar with no days selected

| Before | After |
|--------|-------|
|   screen shot 2018-08-13 at 12 13 01     |   screen shot 2018-08-13 at 12 13 04    |

## Updating the Cypress tests to reflect this use case
- test has been updated to support the no dates selected use case

### DIFF
--- a/web/cypress/integration/main/cal-clear-show-errors.js
+++ b/web/cypress/integration/main/cal-clear-show-errors.js
@@ -20,6 +20,12 @@ context('Calendar Errors clear when selecting incorrect amount of days', () => {
       }
     })
 
+    // submit the form with no dates
+    cy.get('#selectedDays-form').submit({ force: true })
+
+    // trigger the not enough days error
+    cy.get('#submit-error h2 span').should('contain', 'You must select 3 days.')
+
     // select 2 days
     cy.get('.DayPicker-Day[aria-disabled=false]')
       .eq(0)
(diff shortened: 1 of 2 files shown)


pcraig3:
@nmakuch, I think this PR will pass once #353 goes in and this is rebased on top of it.

@timarney, yeah it's not so clear where the change happened. 
This was the PR that fixed it: https://github.com/cds-snc/ircc-rescheduler/pull/283/files

But that logic is still in place. 
The [onSubmit](https://github.com/cds-snc/ircc-rescheduler/blob/master/web/src/pages/CalendarPage.js#L232) call initially calls `validate(values)`, which calls the [Validator](https://github.com/cds-snc/ircc-rescheduler/blob/master/web/src/pages/CalendarPage.js#L158). It's looking for `values` to be `{selectedDays: []}`, but instead it's just passed an empty object (ie, values === `{}`) until the first time the calendar is interacted with. So I'm not sure when that changed. 

If that's ringing any bells, maybe there's another intervention we can make that would make more sense. Else, this seems okay.

timarney:
@pcraig3 ya not sure.

Just merged the other so let's rebase this and should be good.

PR #358, opened 2018-08-13
Boost homepage load time using cache

timarney:
The loading times for slow 3G connections has been quite slow 30+secs.  This in part is due to the bundle size and parse time for the bundle.

Given the the interactivity level and content for the initial landing page we can boost load time but serving up a cached version of the landing page and skip loading the client bundle for the initial page.

# Lighthouse Audit 

**Before:**
before

**After**
after


web/src/server.js
```diff
@@ -129,6 +130,16 @@ server
     res.redirect(`/cancel?language=${language}`)
   })
   .get('/*', async (req, res) => {
+    if (req.url === '/') {
+      let cachedIndex = cache.get('index')
+
+      if (cachedIndex) {
+        return res.locals.redirect
+          ? res.redirect(res.locals.redirect)
+          : res.send(cachedIndex)
```
pcraig3: Reusing this ternary here is a bit awkward, but I get that we don't want to unexpectedly lose redirects on this page.

What do you think of:
 ```
if (cachedIndex && !res.locals.redirect) { 
  return res.send(cachedIndex)
}
```

Comment (timarney) [reply web/src/server.js]: Nice that works better.

**Comment by pcraig3** [comment]
This change means we're always loading the home page without our custom fonts, so they end up flashing in on the second -- register -- page (or if you switch the language or something).

More interestingly, there's something a bit funny going on here. 

When clicking the link to the deployed app (on heroku), it loads the first page in french. Once you continue, it loads the register page in english. 

I'm not getting this behaviour locally, so I'm not sure why -- perhaps it cached the french page the first time around?

Are you seeing the same thing, @timarney?

**Comment by timarney** [comment]
Will get this back up to date with master and than see how we want to move forward.

**Comment by timarney** [comment]
Note: This now has an en + fr cache for the homepage.  Should be all set for further testing or merge.

File: web/src/server.js [hunk 2]
```diff
@@ -160,6 +171,10 @@ server
         document: Document,
       })
 
+      if (req.url === '/') {
+        cache.put(`index_${language}`, html)
```
pcraig3: Is there any reason to invalidate this index at any point?

web/src/server.js
```diff
@@ -145,6 +146,16 @@ server
     res.redirect(`/cancel?language=${language}`)
   })
   .get('/*', async (req, res) => {
+    let language = getStoreCookie(req.cookies, 'language') || 'en'
+    if (req.url === '/') {
```
pcraig3: So I've just been playing around with this, and this `=== '/'` check is pretty strict.

`req.url` includes everything after the domain name, so this will match 
- `rescheduler.cds-snc.ca` 
- `rescheduler.cds-snc.ca/`

But it doesn't match:
- `rescheduler.cds-snc.ca/?language=en`
- `rescheduler.cds-snc.ca/?utm_source=BELA%20email&utm_medium=email`

Means you can get into some weird flows, like:
- load main page (get cached, no JS version)
- press ('français')
- new pageload (`/?language=fr`)
- now you get the JS version of the home page because the url doesn't match anymore
- subsequent language switches will change in-place, but the url will stay the same `/?language=fr`

Not a huge issue, just a bit weird.

More importantly though, we probably want to cache the response for people who come from the BELA email.

web/src/server.js
```diff
@@ -160,6 +171,10 @@ server
         document: Document,
       })
```
pcraig3: This isn't a bug with this code at this point, but setting the language cookie happens in the `render` method.

So, if you were planning to cache URLs even when they include query parameters, you want to be careful.

Specifically,
- if you get the language cookie at the top of the request
- then the cookie changes after running the `render` method
- then you save `cache.put(`index_${language}`, html)`, then you're actually saving the wrong language

(Again, not because your code does this -- only because I was playing around with the code and I got it into this state.)

web/src/server.js
timarney: Bit leary of caching with the params etc...

What I would like to do is get this on staging to get a better idea of overall impact with http2 + gzip.

If we're looking at a difference of 5 secs or something for slow 3g thinking it's worth looking into caching with params etc...

If it's not much ... maybe we removed the caching part and just take the savings from some of the other commits in here for bundle size.

pcraig3: Okay, if you're happy to deploy as-is, I don't see any problems with that.

web/src/server.js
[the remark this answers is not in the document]
pcraig3: Sure thing. Whenever we have changes to make (ie, we redeploy), the cache resets itself anyway, right?

PR #366, opened 2018-08-14
FOUC off!

nmakuch:
## This PR consists of:
*Prior to this pull request, the way we were loading fonts was causing a flash of unstyled content on hard refreshes.* 

*Our styles were being called upon before the fonts were actually being loaded, causing the browser to revert to the back-up font until the proper font is finally loaded*

### Preloading fonts by adding links into the `<head>` of our pages
- `<link rel="preload" href="/fonts/SourceSansPro-Bold.woff" as="font" />` causes the browser to start loading the fonts first, this fixes the FOUC because the files will now be loaded before the reference to the font is called, meaning the back-up font will never have to act as a placeholder
- concept was implemented from this article at [Bramstein](http://www.bramstein.com/writing/preload-hints-for-web-fonts.html)

| Before | After |
|--------|-------|
|   fouc-before   |   fouc-after    |


pcraig3:
Definitely fixes the flash of unstyled content but now the font doesn't load at all for me (trying on the heroku deploy -- I'll try again locally).

| before | after |
|--------|-------|
|  screen shot 2018-08-14 at 3 20 46 pm |  screen shot 2018-08-14 at 3 20 44 pm  |

I am a fan, but 🤷‍♂️

pcraig3:
[There are arguments against inlining fonts](https://www.bramstein.com/writing/web-font-anti-patterns-inlining.html), but I don't think we're going to be too affected by this. 

Our initial HTML request balloons to a larger size, but it'll be smaller once it's gzipped by the production server (heroku doesn't do gzipping for us). 

Do you have any thoughts, @timarney?

timarney:
Visually yes this is better (good connection + newer devices).

Looks like at least on Heroku this seems to add roughly 3 secs to the load time.

And we still get dinged this this error (not too concerned but noting)
screen shot 2018-08-15 at 1 02 10 pm

Given the work we've done to get the bundle size down the fonts being loading in raw are bigger than the bundle now (not due to this PR).

Would be nice to get them loading outside of the bundle to take advantage of HTTP2 which we have on the live server.

Didn't look over these but might be worth looking here:

https://github.com/zachleat/web-font-loading-recipes

nmakuch:
Reverted back to preloading fonts

PR #367, opened 2018-08-15
Use location object for email and phone across the app

timarney:
This updates moves from using hardcoded values for email and phone to use the location file / object.

For now a couple of utility methods (getEmail, getPhone) have been added to that file as well.  Those will get moved out once we determine how we're moving forward with other locations.

### DIFF
--- a/web/src/locations/vancouver.js
+++ b/web/src/locations/vancouver.js
@@ -11,3 +11,21 @@ export const vancouver = {
     dec: ['tues', 'wed'],
   },
 }
+
+export const getEmail = (location = vancouver) => {
+  if (location && location.email) {
+    return location.email
+  }
+}
+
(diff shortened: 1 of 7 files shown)


web/src/locations/vancouver.js
```diff
@@ -11,3 +11,21 @@ export const vancouver = {
...
+  if (location && location.email) {
+    return location.email
+  }
+}
+
+export const getEmailParts = (location = vancouver) => {
+  const email = getEmail(location)
+  let split = email.split('@')
+  return split
+}
```
pcraig3: omg, sorry about this 🙈

timarney: Ha - didn't pick up on what was happening here until the tests failed than notices the split.

PR #369, opened 2018-08-15
Add Sentry notification for missing data redirects

timarney:
Routing users back to fill in missing data shouldn't happen given a typical user flow.  If we end up flagging missing data and redirecting we should be notified we can look further into potential issues.

This update sends a simple message to Sentry for these case so the information is logged and we will receive an email notification.

**Sample of info showing in Sentry:**

screen shot 2018-08-15 at 10 04 49 am

### DIFF
--- a/web/src/server.js
+++ b/web/src/server.js
@@ -72,10 +72,16 @@ server
     const validateCal = new Validator(input, CalendarFields)
 
     if (!validateReg.passes()) {
+      Raven.captureMessage('Missing Data - Register Page', {
+        level: 'info', // one of 'info', 'warning', or 'error'
+      })
       return res.redirect('/register?not-valid=true')
     }
 
     if (!validateCal.passes()) {
+      Raven.captureMessage('Missing Data - Calendar Page', {
+        level: 'info', // one of 'info', 'warning', or 'error'
+      })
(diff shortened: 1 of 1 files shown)


web/src/server.js
```diff
@@ -72,10 +72,16 @@ server
     const validateCal = new Validator(input, CalendarFields)
 
     if (!validateReg.passes()) {
+      Raven.captureMessage('Missing Data - Register Page', {
+        level: 'info', // one of 'info', 'warning', or 'error'
+      })
```
pcraig3: Do you think it's worth being like `'Missing Data - RegisterPage: emailInvalidErrorMessage, paperFileNumberErrorMessage'` or something?

timarney: Thinking given the amount of work it would take to do that might as well.  More data to track down problems = 👍

pcraig3:
One thing I noticed was that all of your data for the registration page looks like it's cleared out if you submit a request with just one field missing.

It's because the `?not-valid=true` query parameter is treated as a form submission on this page.

clearing-fields

Since, under normal circumstances, nobody can get past this page without submitting all the fields, I am happy to release this as-is until we start getting error reports back from Sentry where this is happening.

PR #371, opened 2018-08-15
Check for missing translations on build

pcraig3:
Anik pointed out that a few times we've had English text go out without a French translation. To stop this, we can add a step to the build that will look for missing French translations.

If it's too noisy, or it's blocking us too frequently, we can get rid of it.

Links to documentation in the commit messages.


timarney:
What does  "sourceLocale": "en" do?

pcraig3:
So if you run `yarn compile --strict`, it will fail if we have French translations missing _or_ English translations missing.

By setting a `sourceLocale`, we can get it to ignore missing English translations (because the default is English, so it's fine).

PR #373, opened 2018-08-16
Puppeteer end to end test for our fallback (no js) flow

timarney:
This adds an initial end-to-end for No JS mode

Added two commands for this.

1. Local dev / debugging 
nojs:test - will kick off the no js test in local debug mode.  The will open playback and watch experience with Chromium. 

Sample: https://slack-files.com/T2G2S06PM-FC97YEAHX-b15f2e3775

2. test:full 
We will need to modify our Circle CI steps to kick this off in place of the 'test' command.  We need the build / server to be running for the test to pass.  This test runs in headless mode.  @pcraig3 other than slowing things down because of the build and run do you see any issues with this?

Once this is merged we can start adding more test i.e. checking error messages etc...


web/src/pages/__tests__/NoJS.test.js
```diff
@@ -0,0 +1,122 @@
...
+      let checkedValues = []
+
+      for (let i = 1; i < 4; i++) {
+        let el = `#calendar-checkboxes li:nth-child(${i}) input[type=checkbox]`
+        let checkbox = await page.$(el)
+        await checkbox.click()
+        let checked = await page.evaluate(
+          checkbox => checkbox.checked,
+          checkbox,
+        )
```
pcraig3: Ah nice, we don't have to do any month math here because we always have all the checkboxes in view.

pcraig3:
The code looks pretty straightforward, I think -- very much like our cypress tests, so that's handy.

There's one failure still running this on circleCI (maybe headless browser doesn't have a `close` function or something?)

Other than that, is this good to go or are you still refactoring it or something?

timarney:
K - added a check for the browser close.

Done tweaking this for now.  There's a variable in this to swap to french will probably hook something up so we can make that toggle automated.  It's just in there can as we wanted to do a video capture.

web/package.json
```diff
@@ -10,13 +10,16 @@
...
+    "test:full": "start-server-and-test build:start http://localhost:3004 test",
     "test": "node --icu-data-dir=./node_modules/full-icu node_modules/jest/bin/jest.js",
     "extract": "NODE_ENV=development lingui extract",
     "compile": "NODE_ENV=development lingui compile",
     "add_locale": "lingui add-locale",
     "stats": "BUNDLE_CHECK=true yarn build",
     "a11y:test": "start-server-and-test build:start http://localhost:3004 a11y:run",
     "a11y:run": "node a11y.test.js",
+    "nojs:test": "start-server-and-test build:start http://localhost:3004 nojs:run",
+    "nojs:run": "cross-env NOJS=\"debug\" jest 'NoJS'",
```
pcraig3: Yeah, getting env vars to work is pretty annoying to try to work with otherwise.

pcraig3:
@timarney, sorry I didn't respond earlier. But I don't see any issues, no.

PR #375, opened 2018-08-17
Ensure 3 dates for calendar view

timarney:
This reverts the Calendar tests to bring back some code that was used to ensure that the Calendar tests have enough dates to work with per month view.

These functions got removed during some cleanup work.


pcraig3:
Ah, I thought taking this out was intentional.

timarney:
Out was intentional but it was an oversight as those we're still needed.

PR #376, opened 2018-08-17
Remove screenshot file

timarney:
This removes / ignores a file that crept in with the Puppeteer tests.

Locally I also ran this command
```
git update-index --assume-unchanged [path]
```


pcraig3:
This change doesn't actually remove the screenshot file. There's a command to remove files that git already knows about if you want to keep the file around locally: it's [`git rm --cached {file}`](https://stackoverflow.com/questions/1274057/how-to-make-git-forget-about-a-file-that-was-tracked-but-is-now-in-gitignore).

PR #377, opened 2018-08-17
Only install production dependencies when deploying

pcraig3:
Up to now, we've been installing all production and dev dependencies when deploying the application.

Dev dependencies are used for things like testing or linting, but they aren't needed when running the application. Skipping dev dependencies will make our deployments faster and hopefully get them to actually work again 🤞


pcraig3:
Awesome, let's do it. 👍

Issue #354, opened 2018-08-13
Navigating Back to the Calendar Page


nmakuch:
### New functionality should include:

- New message should appear above the Calendar Container
  - "To change your selections, remove some days first"
- This message should be styled like the Error Messages currently present on the Calendar Page, but without a red border.
- The remove date circles should be black when navigating back to the page.
- The focus of the browser should also be set to the new message when navigating back.

nmakuch:
![](https://github.trello.services/images/mini-trello-icon.png) [Calendar Page functionality while navigating back to page after saving data [S/M]](https://trello.com/c/iWALE79Z/306-calendar-page-functionality-while-navigating-back-to-page-after-saving-data-s-m)

nmakuch:
Closing issue as it was merged into PR #379

PR #379, opened 2018-08-17
Navigating Back to the Calendar Page - Remove Days Message

nmakuch:
*Users were still being confused when navigating back to the Calendar Page. It is apparent that removing days is not as obvious as we hoped at this point in the user flow.*

*A text prompt could solve this issue when navigating back to the cal*
### 1) Adding a text prompt for the user to remove days when navigating back to the calendar after having already saved dates.

| Before | After |
|--------|-------|
|   back2calendar-before     |  back2calendar-after     |

- A daysModified variable has been added to Calendar.js that is set to true whenever a selected day is clicked. This will be used in conjunction with threeDatesArePicked as a switch/trigger for this message.


timarney:
If you rebase this that should fix the tests.

web/src/components/Calendar.js
```diff
@@ -620,6 +635,16 @@ class Calendar extends Component {
...
         </div>
+
+        {this.threeDatesArePicked && this.state.daysModified === false ? (
+          <div className={removeDateMessage}>
+            <h2>
+              <Trans>To change your selections, remove some days first</Trans>
+            </h2>
+          </div>
+        ) : (
+          ''
```
timarney: Better to return null here vs ''

nmakuch:
that's a good point, didn't think about the error being triggered on top of the prompt

nmakuch:
## Update

*Didn't take into account that the error message could be triggered on top of this prompt*

*Error message and Prompt had duplicate language. The solution was splitting this message apart while the prompt was on screen*

### 1) Split Error Message apart while Remove Day Prompt is on screen
- While the prompt is present on screen, the messages will now be split as in the screen shot below:
screen shot 2018-08-17 at 14 45 58

- Once the date has been removed and the daysModified boolean switches to true, the message will go back to normal:
screen shot 2018-08-17 at 14 47 53

web/src/components/Calendar.js
```diff
@@ -522,18 +532,34 @@ class Calendar extends Component {
...
+        if (this.threeDatesArePicked && !this.state.daysModified) {
+          await this.setState({
+            errorMessage: (
+              <Trans>You can&rsquo;t select more than 3 days.</Trans>
+            ),
+          })
+
+          this.errorContainer.focus()
+          return
+        }
```
pcraig3: Returning here would mean that we would lose the `logEvent` call below. 

If you just set the message in an if/else, then you could let it fell through.

maybe more like this:

```js
if (selectedDays.length >= dayLimit) {
  let errorMessage

  if (this.threeDatesArePicked && !this.state.daysModified) {
    errorMessage = <Trans>You can&rsquo;t select more than 3 days.
  } else {
    errorMessage = <Trans>You can&rsquo;t select more than 3 days. To change your selections, remove some days first.
  }

 await this.setState({ errorMessage })
 this.errorContainer.focus()
 logEvent('Calendar', 'Select', 'Error: More than 3 days')
 return
}
```

web/src/components/Calendar.js
```diff
@@ -542,6 +568,7 @@ class Calendar extends Component {
             </Trans>
           ),
         })
+
```
pcraig3: Also, I know that's not technically part of this PR, but see if the focus is working properly on safari.

Otherwise, we could add a `scrollTo` command like we do here: https://github.com/cds-snc/ircc-rescheduler/blob/cdc7ab26b56cb523f8443a52d5ebf36f44e2696f/web/src/pages/CalendarPage.js#L252

pcraig3: Yeah, it's a bit janky on my phone (safari on iOS). Let's throw that in and see if it improves.

nmakuch:
## UPDATE:

*The update to the Error Message + Prompt functionality is a result of a chat me and Eman had. I think it makes more sense than having the two separate messages on screen. Open to discussion if needed:*

### 1) Merging the Remove Day Prompt into the Error Message when an error is triggered

back2calendar-newerror

### 2) Changing focus to be set to Remove Day Prompt when navigating back to the Calendar Page

back2calendar-focus

### This PR addresses [Issue 354](https://github.com/cds-snc/ircc-rescheduler/issues/354)

pcraig3:
Just did another change.

- fixed focus ring
- added period
- removed `scrollTo` when coming back to calendar page
  - (seemed a bit on the nose)
  - we can put it back if we want it

| before | after |
|--------|-------|
| screen shot 2018-08-21 at 6 28 25 pm  | screen shot 2018-08-21 at 6 28 38 pm   |

pcraig3:
If we do want the `scrollTo` back, let's put an offset in there to make it nicer.

`window.scrollTo(0, this.removeDateContainer.offsetTop - 30)`

nmakuch:
Looks really good, that offset property is super useful

PR #380, opened 2018-08-17
WIP Return validate

timarney:
## User-visible error message when missing data in email requests
This is setup but we need feedback if we want a custom error message for this use case.

i.e. **Some information is missing** ... might want to add note about feedback link or something
https://ircc-development-pr-380.herokuapp.com/register?not-valid=true

cc: @zakkain @emanelfy 

## Auto submit form if not-valid found
If not-valid is found auto-submit the form to show error messages

## Add custom general error message for not valid
In cases where not valid = true add custom message to better describe the problem


zakkain:
sorry can you explain this a bit more? The form auto-submits when what is not-valid?

timarney:
@zakkain based on 5 whys: retro we are looking to display a better error message.  This would be part of the "Make the user feedback mechanism more obvious" outcome.

Given a use case where for some reason the user has been able to get to the review page with missing information.  The server validates the data and redirects back to the page that is missing the information so say the Registration page (missing fullName).   The url in this case would be /register?not-valid=true.  For the code we know that no-valid is true so we're not in a typical user flow.  In which case we will auto-validate to show the default error messages i.e. your missing fullName ...

Previously we redirected to the page but didn't auto validate or show any messages so the user was left in basically a start again state but possibly thinking that had submitted the data and all was good.

We're now able to redirect and show some sort of message saying roughly something messed up... retry or contact us.

pcraig3:
From a dev perspective, it's always better to use the thing that already works than come up with a new special case. Especially when that case isn't _supposed_ to be able to happen.

timarney:
## Register (missing data)
https://ircc-development-pr-380.herokuapp.com/register?not-valid=true
## Calendar (missing data)
https://ircc-development-pr-380.herokuapp.com/calendar?not-valid=true

@pcraig3 thanks for you work on the branch deploys.  This PR is a shining example of the need.

web/src/pages/CalendarPage.js
```diff
@@ -196,6 +198,19 @@ class CalendarPage extends Component {
...
   }
 
+  /* 
+  If the user has pressed submit we want to ignore the
+  not-valid param i.e. we're in a regular 
+  form submit scenario
+  */
+
+  hasNotValid() {
+    if (this.state.submitClicked) {
```
pcraig3: This is never being set in this file. Are we wanting to set it to "true" when the submit button is clicked?

pcraig3: Otherwise, we could just remove the check?

web/src/pages/RegistrationPage.js
```diff
@@ -137,19 +137,41 @@ class RegistrationPage extends React.Component {
...
+  /* 
+  If the user has pressed submit we want to ignore the
+  not-valid param i.e. we're in a regular 
+  form submit scenario
+  */
+
+  hasNotValid() {
+    if (this.state.submitClicked) {
+      return false
+    }
```
pcraig3: Do we need this check? What happens (or could happen) if we don't have it?

web/src/pages/RegistrationPage.js
```diff
@@ -193,14 +215,24 @@ class RegistrationPage extends React.Component {
...
+              Object.keys(errorsNoJS).length && !submitError
+                ? generalMessage
+                : submitError
             return (
-              <form id="register-form" onSubmit={handleSubmit}>
+              <form
+                id="register-form"
+                ref={el => {
+                  if (!this.form && notValid) {
+                    el.dispatchEvent(new Event('submit')) // eslint-disable-line no-undef
```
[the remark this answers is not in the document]
timarney: It's set on the line after the dispatchEvent.  Could be moved up above the dispatch.

```
if (!this.form && notValid) {
                    el.dispatchEvent(new Event('submit')) // eslint-disable-line no-undef
                    this.form = el
                  }
```

timarney:
Going to clean this up further.

Originally I had different error messages showing or at least the capability to show different error messages for a not-valid on load vs a form submit where the not-valid still exists in the url.

Given we're showing the same error messages for both states now ... this can be cleaned up further.

pcraig3:
@timarney, FYI, the reason for this is probably the same thing we saw before when all the BELA params were being processed on the register page even though they came from the `/` root url. That's why we threw [this extra check in](https://github.com/cds-snc/ircc-rescheduler/blob/73868ab71a676a8887afc6521dd23749bb1f13e1/web/src/pages/RegistrationPage.js#L187), because the `location` param can be behind by one page. (probably a server/client difference.)

PR #385, opened 2018-08-20
Ignore Puppeteer tests in `yarn test`

pcraig3:
Since adding the puppeteer tests, our `yarn test` command would fail when running without building and starting the app first.

By ignoring the NoJS.test.js file, we can run the rest of our jest tests from the command line without the faff of spinning the app up and down.

This deceptively simple problem turned out to be a bit more complicated than just doing a regex or passing a new command line argument.

In the end, I created a dedicated `jest.config.js` file that jest will use on CI test runs, and a second config file that ignores the NoJS tests so that we can run the other ones locally. Using the spread operator to reuse nearly the same config options between both files.

File matching using regex is also still working as before. ie, if I just wanted to run calendar-related tests, I can still use `yarn test Calendar`.

More info on jest configuration can be found here:
- https://jestjs.io/docs/en/configuration.html

### DIFF
--- a/web/jest-no-js.config.js
+++ b/web/jest-no-js.config.js
@@ -1,0 +1,8 @@
+'use strict'
+
+const jestDefaultConfig = require('./jest.config.js')
+
+module.exports = {
+  ...jestDefaultConfig,
+  testPathIgnorePatterns: ['/node_modules/', 'NoJS.test.js'],
+}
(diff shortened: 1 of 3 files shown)


timarney:
This is great.  This puts us back to a nice place where we only need to spin up the server in cases where the no js tests need to run.

PR #386, opened 2018-08-20
update deps

timarney:
Updates the app to use the latest dependancies.

Also notes a "bug" where in order for package.json to save the --latest flag needs to be passed to the yarn command.

```
 yarn upgrade-interactive --latest
```

Given we're not pinning version numbers at with point should be fine using the --latest flag (for now).


timarney:
Will make a separate PR tomorrow for SNYK.

PR #388, opened 2018-08-22
[WIP] Add `location` global field

pcraig3:
Note: this is a *WIP* PR while the tests are b0rked but it's pretty well finished.

***

This PR changes our internal data store structure a bit:

- adds a new `GLOBALS` key for global value (like language, and now location)
  - (Incidentally, the definition of a global value is that it can be set from any page)
- Adds `location` as a global value
- Lets us set multiple global values at once
- Explicitly ignores extra params we don't expect
  - ie, this will now work: 
  - `/?utm_source=BELA%20email&utm_medium=email&location=montreal&language=fr`

This PR does not:
- hook up any interface to `location` -- it can only be set via a query param at present
- use `location` for anything -- we're just setting a value

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -41,6 +41,7 @@
     "body-parser": "^1.18.3",
     "cookie-parser": "^1.4.3",
     "date-fns": "^1.29.0",
+    "deepmerge": "^2.1.1",
     "emotion": "^9.2.6",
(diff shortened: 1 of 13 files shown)


pcraig3:
Closing this since #392 saves our location not using subdomains rather than query params + cookies.

PR #389, opened 2018-08-23
Add a check for window before using scrollTo

timarney:
Not all environments have access to the window object.

Our unit tests for instance don't have access.

This fixes errors being thrown in the tests
```
console.error node_modules/jsdom/lib/jsdom/virtual-console.js:29
    Error: Not implemented: window.scrollTo
```

### DIFF
--- a/web/src/components/Calendar.js
+++ b/web/src/components/Calendar.js
@@ -21,6 +21,7 @@ import {
 } from '../utils/calendarDates'
 import parse from 'date-fns/parse'
 import { logEvent } from '../utils/analytics'
+import { windowExists } from '../utils/windowExists'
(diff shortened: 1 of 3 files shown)


web/src/components/Calendar.js
```diff
@@ -506,7 +507,6 @@ class Calendar extends Component {
 
   componentDidMount() {
     if (this.threeDatesArePicked && this.props.input.value.length === 3) {
-      // window.scrollTo(0, this.removeDateContainer.offsetTop)
```
pcraig3: Good catch -- I left this in the other day but if we're not using it, we want it gone.

PR #390, opened 2018-08-23
Subdomains

timarney:
## Initial Subdomain setup

Sets up the server to handle locations via subdomain.

For local testing use vcap.me:3004 i.e. http://mtrl.vcap.me:3004

Valid subdomains have been whitelisted.

```
const domainOptions = { namespace: '', whitelist: ['van', 'cal', 'mtrl'] }
```

1. If a subdomain is found but is not on the list it will redirect.
2. If no subdomain is found for now it will default to Vancouver.

### DIFF
--- a/web/src/server.js
+++ b/web/src/server.js
@@ -83,7 +83,7 @@ const getPrimarySubdomain = function(req, res, next) {
 
   if (
     !domainOptions.whitelist.includes(req.subdomain) &&
-    req.url !== '/not-found'
+    req.url.indexOf('not-found') === -1
   ) {
     // redirect to generic not found page
     /*
@@ -91,8 +91,10 @@ const getPrimarySubdomain = function(req, res, next) {
     so they don't end up at the wrong location 
     */
 
-    // Todo: Would rather not introduce another env variable here  
-    return res.redirect(process.env.SITE_URL + '/not-found')
+    // Todo: Would rather not introduce another env variable here
+    // hard code for testing
+    return res.redirect('https://rescheduler-dev.cds-snc.ca/not-found')
+    //return res.redirect(process.env.SITE_URL + '/not-found')
   }
 
   next()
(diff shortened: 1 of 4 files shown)


web/src/server.js
```diff
@@ -70,13 +71,44 @@ const captureMessage = (title = '', validate) => {
   })
 }
 
+const domainOptions = { namespace: '', whitelist: ['van', 'cal', 'mtrl'] }
```
pcraig3: Can we make these match our location files (ie, `vancouver`, `calgary`)?

Or is there a reason that they're different?

web/src/server.js
```diff
@@ -70,13 +71,44 @@ const captureMessage = (title = '', validate) => {
...
+
+const getPrimarySubdomain = function(req, res, next) {
+  req.subdomain = req.subdomains.slice(-1).pop()
```
pcraig3: Is req.subdomain a value that express uses, or is that something that we're setting?

web/src/server.js
timarney: For sure.

web/src/server.js
timarney: Adding that.  Can call it something else to make it more obvious.  As far as I know that's a common way to handle.

```
.use(getPrimarySubdomain)
```

pcraig3: No, it's okay -- just wondering. 

When you try to set `req.subdomains` (for example, `req.subdomains = req.subdomains.slice(-1)`), you get a big fat error message because we're not allowed to write to it.

pcraig3: So this means you can do something like `vancouver.some.other.subdomains.rescheduler.cds-snc.ca`, right? 

Bit weird, but seems like we can live with it. As long as the first one is a valid location, you're in the clear and the app won't actually do anything different.

timarney: Correct - agreed it's weird but it's valid.

PR #391, opened 2018-08-24
Adds ability to have blocked days on the calendar

timarney:
This adds a 'blocked' property to the location object to block off calendar days

export const test_location = {
  recurring: {
    ...
  },
  blocked: '2018-10-02, 2018-10-03, 2018-11-21', // use CSV format => 2018-10-02, 2018-10-03, 2018-11-21
}

If the property exists the CSV string will be parsed and dates blocked on those values
<hr>
Adds utility method to parse a CSV string of dates

dateSetFromString takes a 'user entered' date string  i.e. 2018-10-17, 2018-11-06, 2018-11-07
and creates a new JavaScript set.

This will be used for checking if a given date is in the list set.has('2018-10-17')

### DIFF
--- a/web/src/utils/calendarDates.js
+++ b/web/src/utils/calendarDates.js
@@ -87,11 +87,7 @@ export const firstValidDay = (location = vancouver, date) => {
   return parse(addDays(date, i))
 }
 
-const isValidDayForLocation = (
-  location = {},
-  month = '',
-  date = new Date(),
-) => {
+const isValidDayForLocation = (location, month = '', date = new Date()) => {
   // eslint-disable-next-line security/detect-object-injection
   if (location && location.recurring[month]) {
     const result = checkLocationDays(location, month, date)
@@ -168,6 +164,9 @@ const getAllowedDays = (
 
 export const checkLocationDays = (location, month, date) => {
   let valid = false
+
+  const dateSet = dateSetFromString(location.blocked)
+
   // eslint-disable-next-line security/detect-object-injection
   location.recurring[month].forEach(day => {
     /* 
@@ -175,8 +174,11 @@ export const checkLocationDays = (location, month, date) => {
       isMonday, isTuesday
     */
     const result = isDayValid(day, date)
+    const dateFormatted = format(date, 'YYYY-MM-DD')
 
-    if (result.valid) {
+    // checks to see if a day is valid and that it's not blocked
+    // weekends are blocked automatically
+    if (result.valid && !dateSet.has(dateFormatted)) {
       valid = result.valid
     }
   })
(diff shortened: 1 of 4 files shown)


web/src/utils/calendarDates.js
```diff
@@ -304,3 +315,19 @@ export const getShortMonthName = (date = new Date()) => {
...
   return format(date, 'YYYY-MM-DD')
 }
+
+/*--------------------------------------------*
+ * Dates from CSV string
+ * '2018-10-17, 2018-11-06, 2018-11-07'
+ *--------------------------------------------*/
+
+export const dateSetFromString = dates => {
+  if (!dates) return new Set()
```
pcraig3: Oh nice, I've wanted to do one-line returns several times but I didn't know this syntax would work.

web/src/locations/test_location.js
```diff
@@ -0,0 +1,14 @@
...
+    jun: ['wed', 'thurs'],
+    jul: ['wed', 'thurs'],
+    aug: ['wed', 'thurs'],
+    sep: ['wed', 'thurs'],
+    oct: ['tues', 'wed'],
+    nov: ['tues', 'wed'],
+    dec: ['tues', 'wed'],
+  },
+  blocked: '2018-10-02, 2018-10-03, 2018-11-21', // use CSV format => 2018-10-02, 2018-10-03, 2018-11-21
+}
```
pcraig3: Do we need this extra file, or can we get away with not having it? 

It looks like you're mocking it out in the jest test where you need it.

web/src/utils/calendarDates.js
```diff
@@ -167,16 +167,27 @@ const getAllowedDays = (
...
 export const checkLocationDays = (location, month, date) => {
+
   let valid = false
+
+  location.dateSet = new Set() // default to an empty set
+
+  if (location.blocked) {
+    location.dateSet = dateSetFromString(location.blocked)
+  }
+
```
pcraig3: since `dateSetFromString` returns an empty Set if `location.blocked` is falsey (ie, undefined), you don't even need this `if` statement or the empty set initially, right? You could just do line 176 and be done with it.

pcraig3: Actually, could we assign this to just a variable called `blockedDays` or something? 
Correct me if I'm wrong, but since `location` is an object, this modifies the original value we pass in, right?

web/src/locations/test_location.js
timarney: Right - can remove now.

web/src/utils/calendarDates.js
timarney: Correct.  Nice catch.

PR #392, opened 2018-08-24
Set and cache a global location in the app

pcraig3:
# This PR imports a location into the app.

- On the server, we have a `location` string which we will get from a subdomain (for now this is hard-coded)
- using the string, we `require` a location file
  - if the location file doesn't exist, we set a placeholder object (but maybe it should be false)
- the `locationString` is available on the server and on the client for the initial render
- in each environment, we set a location, cache it, and pass it to React context
  - this means that it's available to us if we call `getGlocalLocation`, and it's also available as `this.props.context.location` in our page components

## questions

(@timarney, have a look through these)

- If we don't find a location file, I'm returning `{ id: ${locationString} }` as the placeholder location. Should this be more explicit about being empty?
- For tests, I'm defaulting location to `vancouver`. Seems okay, right? Though of having an "only for tests" location config, but figured we'd rather test the real thing.
- I'm throwing an `Error` if we are trying to load a location the first time but we haven't passed through a location string. Should this be handled differently?

## More details

Passing the location object through the context is great, but a lot of our date-calculating code needs to know about the location outside of React-land, so we need to be able to serve up the location somewhere else.

Wrote `LocationCache` as a small 'set once and cache' JS module (using the module pattern: https://toddmotto.com/mastering-the-module-pattern/) that will save a location once and then return the saved location thereafter.

This means we need to call `getLocation(locationString)` (a) once on the server, and then (b) once on the client.

Anything passed through `getInitialProps` is available (a) on the server, and then (b) once on the client.
(ie, `props.context` is available on the first page render, but then we don't have it anymore as we move between pages).

This means we can pass a subdomain string through from the server and it will be available on the server and the client. Once `LocationCache` is called once on each environment, it saves the location and we don't need to reset it again.

So we never need to inspect the subdomain path on the client, nor do we need to attach a global variable to window or anything like that.


web/src/locations/index.js
```diff
@@ -0,0 +1,66 @@
...
+          - no previous location has been saved
+          - no location string provided
+        throw an error
+      */
+      if (!location) {
+        if (process.env.NODE_ENV === 'test') {
+          location = 'vancouver'
+        } else {
+          throw new Error(
+            'LocationCache.getLocation: `location` must be a non-empty string',
```
timarney: Can test but do you know if this catches with Sentry - thinking it would.

web/src/locations/index.js
```diff
@@ -0,0 +1,66 @@
+const LocationCache = (function() {
+  let _cachedLocation = undefined
+
+  /* this is called once on the server and once on the client */
+  const _setLocation = location => {
+    try {
+      // eslint-disable-next-line security/detect-non-literal-require
+      _cachedLocation = require(`../locations/${location}.js`)
```
timarney: This is nice with the catch vs scanning the directory

timarney:
For now - For test defaulting to Vancouver seems fine.  There's a use can in the blocked days PR where I'm passing in a test location file as I need to ensure the Calendar view + which days are blocked.  Probably some more use cases will come up but defaulting is fine. 

RE:
If we don't find a location file, I'm returning { id: ${locationString} } as the placeholder location. Should this be more explicit about being empty?

Thinking we should do something else here but I'm not sure what at this point.  Let's do what you have for now.

RE:  if we are trying to load a location the first time but we haven't passed through a location string. 

Seems at this point throwing makes the most sense to me for this case.

pcraig3:
> Could add a meta description 

Yup, I'll go in and add a `<meta>` tag

> do you know if this catches with Sentry - thinking it would

Catches sometimes for sure, but I'll test it out a bit more


As to your other comments, I'm inclined to agree with the "address details later but get this in sooner" approach.

pcraig3:
Okay, @timarney, a bit more feedback.

1. The error in the `LocationCache` causes a hard tailspin if it gets thrown (in my case, I was deliberately triggering one). My suspicion is that this is because `withProvider` wraps every page, including the error page. So the url changes to `/error` but the page either goes blank or shows a stacktrace, depending on if it happens on the sever or the client. 
1.b. Note that this isn't possible 🤞🤞🤞 so I don't think we'll see it happen, but it should be addressed

2. The cache on the homepage will cache the first location it gets. This means that wherever the first person who sees it visits from, it will save that location in the static HTML on the first page. Subsequent pages will have the real location. 
2.b. since our home page makes no distinction between locations on the home page, we can get away with this for now.

timarney:
Okay on #1 + 2.

#2 thinking this would be fine anyway with the sub-domains.

web/src/components/LanguageSwitcher.js
```diff
@@ -59,12 +59,15 @@ class LanguageSwitcher extends React.Component {
...
-    let { context: { setStore } = {} } = this.props
+    let {
+      context: { setStore, location: { id: locationId = 'not set' } = {} } = {},
+    } = this.props
 
     return (
       <form>
         <Helmet>
           <html lang={this.state.language} />
+          <meta name="keywords" content={`location=${locationId}`} />
```
timarney: That works

web/cypress/integration/main/cal-introtext.spec.js
```diff
@@ -0,0 +1,29 @@
+/* When toggling the language the intro text Month name should be in the correct language. */
+import { getStartMonthName } from '../../../src/utils/calendarDates'
+import { getGlobalLocation } from '../../../src/locations'
+
+context('Calendar intro text', () => {
+  it('should show intro text in the correct language', () => {
+    cy.visit('/calendar')
+    cy.url().should('contain', '/calendar')
+
+    /*
```
timarney: Perfect.

PR #396, opened 2018-08-24
Supress jsdom console.error output

sleepycat:
This is a workaround for jsdom/jsdom#1422. This just prevents a bunch of
console.error output from cluttering the test results.


pcraig3:
Yeah, this is perfect. Thanks!

PR #402, opened 2018-08-27
Lingui upgrade

timarney:
lingui-react has been deprecated and is now @lingui/react instead.

This PR follows the migration for the core packages.
https://lingui.js.org/releases/migration-2.html

Tests have also been updated.

## Note 
Default wrapping components removed in this package update.

This mean no more "extra" span tags.

There should be some further CSS clean up that can be done but @nmakuch might be better to tackle that further i.e. 

Maybe?
```
const button = css`
  ...
  span {
    line-height: 1;
  }
`
```

@pcraig3 I opted to add an option to the config to keep the message.json rougly as is.

By default now the format is minimal which sounds like it would end up with no line #'s etc... just simple JSON only.  

Thoughts on moving to the simple JSON format.  Was thinking maybe land this first and than try it out.

https://lingui.js.org/releases/migration-2.html#default-message-catalog-format
```
"@lingui": {
    "srcPathIgnorePatterns": [
      "/node_modules/",
      "build"
    ],
    "sourceLocale": "en",
    "format": "lingui"
  }

```


pcraig3:
Noticed a couple visual changes but they seem pretty small so I'm gonna have a pop at them.

timarney:
There was one tweak I made for the Start Button



Prior we had a JSX space which was adding a HTML comment in turn causing the tests to fail.

PR #404, opened 2018-08-27
WIP - Family Option

nmakuch:


pcraig3:
Sorry, I was on a bus for ages, so I didn't have a change to post this earlier.

### Adding a checkbox and a textarea. 

Currently, they are nested inside each other in a weird way, but you're right to pull them out of each other. Having a label in a label seems like a bad idea.

//

The validation rules that exist now are that one is required if the other exists. This leads to a weird interaction pattern where
- if you click to reschedule your family
- if you start writing in the text area
- if you then unclick to reschedule your family

then you get a validation error and it's hard for you to clear the textarea (you have to click, delete, unclick again).

A better way forward would be:
- checkbox is always optional
- the textarea is required if the checkbox is clicked (this is what it does currently)

In this case, we can just save whatever is in the textarea regardless, but only show the family member names on the review page if the checkbox has been checked.

If we only disable the textarea when JS loads, then these existing validation rules will work just fine.

//

Definitely we want to add this to the email to staff. 
Not sure if we want to add this to the email to users.

(Zak would be the one to ask).

// 

No tests or anything exist at present (of course).

//

And, as a final note, we talked about this yesterday, but just to reiterate.

Final form treats checkbox inputs as arrays. ([As per this example here](https://codesandbox.io/s/ww40y2m595).)

This means that it expects the initial value passed in to be `{ familyCheck: ['familyCheck'] }`, and, based on the url string, it's getting `{ familyCheck: 'familyCheck' }`.

[I earlier added a function](https://github.com/cds-snc/ircc-rescheduler/pull/404/commits/d6803e44040a1e34ff1f8bbab8cf4447a903aebf) to work around some of the weirdness (ie, on pageload `familyCheck` isn't in values, but then once you check and uncheck it keeps around the key pointing at an empty array), so this is another one that I don't have a good generic answer for. 

It's not super hard to solve the problem for this one case though since we know what we're looking for.

timarney:
Nice work on that last set of commit @pcraig3 🥇 

One last thing from what I see (outside of removing the debug on screen text)

I'm thinking we should either clear the family option text if you toggle the checkbox or clear it after submit if the checkbox isn't checked.

Currently the value for the family option will show on the review page without the checkbox being set.

## Toggling the checkbox keeps the text
screen shot 2018-09-05 at 9 32 40 am

## Review Page
<hr>
screen shot 2018-09-05 at 9 34 26 am

timarney:
Almost there ...

1. When not entering other family members we end up with a trailing comma on the "Individuals to be rescheduled:"
screen shot 2018-09-06 at 2 16 04 pm

2. In the NoJs flow if I enter text in the family members text area but I don't check the box I don't get flagged that I need to check the box (looking thoughts on this?)

pcraig3:
> When not entering other family members we end up with a trailing comma on the "Individuals to be rescheduled

Good point, will remove this.

> In the NoJs flow if I enter text in the family members text area but I don't check the box I don't get flagged that I need to check the box (looking thoughts on this?)

Yikes, I forgot about this. 
If the checkbox is not checked, the field will be cleared once it's submitted, but that's not a very fault-tolerant approach.

I haven't thought too much about this, no.

PR #405, opened 2018-08-28
Allow locations to have custom receiving address

timarney:
The app has been using an env variable for the receiving address.  Now that we're moving to multiple locations we'll need the ability to have a receiving address per location.

This adds a receiving address prop to the location object and a generic lookup function.

Ensure location and email checks have also been added.  Note: Additional logic should be added via another PR for proper crash /error page handling.


pcraig3:
Once this is deployed, our staging site will start sending all of its emails to the real Vancouver address. 

Since we probably don't want to do this, we can keep the `RAZZLE_IRCC_RECEIVING_ADDRESS` email there and then inspect `RAZZLE_STAGE` before actually sending the email. 

How does that strike you? 

We would have to keep the env var around + rewrite the docs, but it would stop us sending tons of test emails to the wrong place.

timarney:
Good point.

So are you thinking like this?

```
export const getReceivingEmail = (location = getGlobalLocation()) => {
if(RAZZLE_STAGE ){
 return RAZZLE_IRCC_RECEIVING_ADDRESS
}


if (location && location.receivingEmail) {
    return location.receivingEmail
  }
   throw new Error('Receiving address is not defined')
}
```

pcraig3:
That's better than what I was thinking, actually. 

I was thinking `receivingAddress = RAZZLE_STAGE === production ? getReceivingEmail() : RAZZLE_IRCC_RECEIVING_ADDRESS`, but yours is much more contained.

Although, it raises the problem we could throw an error in production but not in staging. (ie, if staging never checks for the `location.receivingEmail`, it would never throw an error until it hit production.

web/src/locations/index.js
```diff
@@ -67,6 +67,10 @@ export const getPhone = (location = getGlobalLocation()) => {
 }
 
 export const getReceivingEmail = (location = getGlobalLocation()) => {
+  if (process.env.RAZZLE_STAGE) {
+    return process.env.RAZZLE_IRCC_RECEIVING_ADDRESS
```
pcraig3: What if we rename this to `RAZZLE_TEST_RECEIVING_ADDRESS` or something?

app.json
```diff
@@ -6,7 +6,6 @@
     "RAZZLE_AWS_ACCESS_KEY_ID": { "required": true },
     "RAZZLE_AWS_REGION": { "required": true },
     "RAZZLE_AWS_SECRET_ACCESS_KEY": { "required": true },
-    "RAZZLE_IRCC_RECEIVING_ADDRESS": { "required": true },
```
pcraig3: Should change the name of this rather than outright removing it, I think.

timarney: 👍 Missed that one as I was reverting some of the files.

pcraig3:
So, what was happening was that the new env var wasn't set in CircleCI, so the app wasn't starting up.

Instead, it was throwing an error, trying to load the error page, but then throwing the error again -- into infinity. Logged over 5000 errors in Sentry 😓   

screen shot 2018-08-28 at 5 00 11 pm

PR #406, opened 2018-08-28
Use env var for site url / redirect

timarney:
This update makes use of the existing RAZZLE_SITE_URL env variable.  We'll need to update that to allow the site to redirect to a "generic" not found and maybe error page as needed.

@pcraig3 to be discussed should we keep the "base domain" alive to host generic pages (no location etc... needed)?

### DIFF
--- a/app.json
+++ b/app.json
@@ -9,7 +9,8 @@
     "RAZZLE_IRCC_RECEIVING_ADDRESS": { "required": true },
     "RAZZLE_PAPER_FILE_NUMBER_PATTERN": { "required": true },
     "RAZZLE_PUBLIC_DIR": { "required": true },
-    "RAZZLE_SENDING_ADDRESS": { "required": true }
+    "RAZZLE_SENDING_ADDRESS": { "required": true },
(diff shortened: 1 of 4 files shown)


pcraig3:
Grr, all of our heroku deploys are dependent on the PR number.

ie → https://ircc-development-pr-406.herokuapp.com/

If we don't want to add some logic to try and treat the heroku config differently (which _I_ don't think we should spend a lot of time with personally), we could just hardcode this into Heroku as `rescheduler-dev` or something?

pcraig3:
Question: does the site URL look like?

- `https://rescheduler-dev.cds-snc.ca`
- `rescheduler-dev.cds-snc.ca`

If we do the first one, it's hard for us to add locations in.
If we do the second one, we have to add in the `https` (or not, if we're on localhost) ourselves. 

I'm thinking option 2.

timarney:
Ya let's go with option 2

pcraig3:
Cool, updated in Heroku as well. 

image

I don't think this will get our heroku deploys to actually work again, but at least they won't immediately fail to build.

PR #407, opened 2018-08-28
Use http as redirect protocol if running locally

pcraig3:
So this is a best-guess.

If we're running locally, we're probably on 3004.
If that's in the site URL, we want to stay on HTTP.

**Other option that occurred to me:**
Might be better to re-use the `RAZZLE_COOKIE_HTTP` var since that one is only for HTTP environments.

### DIFF
--- a/web/README.md
+++ b/web/README.md
@@ -51,12 +51,12 @@ RAZZLE_STAGE='development'
 
 These options are set locally, whether running in development (`yarn dev`) or production (`yarn start`) modes. They are never deployed to the server.
 
-- `RAZZLE_COOKIE_HTTP`: Secure cookies will only served over `https` connections. Since we develop locally on `http`, we need this variable set to `true` so that we can save data between pages.
+- `RAZZLE_IS_HTTP`: Secure cookies will only served over `https` connections. Similarly, when we do redirects on the server and we need the full domain string, we need to know if we're on `https` or `http`. Since we develop locally on `http`, but run our production site on `https`, we need this variable set to `true` so that we can save data between pages and redirect to the right address.
 
 ##### sample `web/.env.local` file
 
 ```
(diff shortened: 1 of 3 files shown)


timarney:
Maybe RAZZLE_COOKIE_HTTP should become RAZZLE_USE_HTTPS, RAZZLE_IS_HTTPS  or RAZZLE_HTTPS.  Something to make it more generic so it re-usable in multiple spots.

If not thinking this would be better as a little wrapper function in case we need it elsewhere.

```
useHttps() => {
 return  process.env.RAZZLE_SITE_URL.endsWith('3004')
      ? 'http'
      : 'https'
}
```

pcraig3:
`process.env.RAZZLE_IS_HTTP` is a pretty easy boolean check on its own. Since we're not doing the 'https' protocol anywhere else (yet), I'm inclined to leave this as is.

web/README.md
```diff
@@ -51,12 +51,12 @@ RAZZLE_STAGE='development'
 
 These options are set locally, whether running in development (`yarn dev`) or production (`yarn start`) modes. They are never deployed to the server.
 
-- `RAZZLE_COOKIE_HTTP`: Secure cookies will only served over `https` connections. Since we develop locally on `http`, we need this variable set to `true` so that we can save data between pages.
+- `RAZZLE_IS_HTTP`: Secure cookies will only served over `https` connections. Similarly, when we do directs on the server and we need the full domain string, we need to know if we're on `https` or `http`. Since we develop locally on `http`, but run our production site on `https`, we need this variable set to `true` so that we can save data between pages and redirect to the right address.
```
timarney: do redirects ya?

pcraig3: Yes 👍

PR #408, opened 2018-08-28
Stop infinite loop for startup errors

pcraig3:
In our middleware, we will throw an error if there is no location string or if there is no receiving email address shown.

On the landing page, if we hit an error, it would redirect to the error page.
Unfortunately, the error page would also encounter the same error.

This meant that we would end up with an infinite loop if we ran into any of these.

This commit generalises some of the code, and helps mitigate infinite error loops. **Line 100 is the operative one**: `if (req.path === '/error') return next()`

*Note: this doesn't get us out of the woods because we can still have the location string taking down the app if it's not there at all, but it helps.*

### DIFF
--- a/web/src/locations/index.js
+++ b/web/src/locations/index.js
@@ -75,7 +75,7 @@ export const getReceivingEmail = (location = getGlobalLocation()) => {
     return process.env.RAZZLE_IRCC_TEST_RECEIVING_ADDRESS
   }
 
-  if (location && location.receivingEmail && !process.env.RAZZLE_STAGE) {
+  if (location && location.receivingEmail) {
     return location.receivingEmail
   }
 
--- a/web/src/email/sendmail.js
+++ b/web/src/email/sendmail.js
@@ -14,6 +14,16 @@ if (process.env.NODE_ENV !== 'test') {
     throw new Error('process.env.RAZZLE_SENDING_ADDRESS was not found')
   if (!process.env.RAZZLE_SITE_URL)
     throw new Error('process.env.RAZZLE_SITE_URL was not found')
+  if (
+    process.env.RAZZLE_STAGE &&
+    process.env.RAZZLE_STAGE !== 'production' &&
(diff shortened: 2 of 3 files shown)


timarney:
This will help. 

As part of this...

Debating whether we should have the RAZZLE_IRCC_RECEIVING_ADDRESS as part of the enforced sendmail script to shut things down sooner.

Like we do for the other vars:
```
if (!process.env.RAZZLE_SENDING_ADDRESS)
    throw new Error('process.env.RAZZLE_SENDING_ADDRESS was not found')
```

Maybe wrapped with with check for RAZZLE_STAGE.

timarney:
FYI - setting up a new page to redirect to 
```
return res.redirect('/500')
```

web/src/locations/index.js
```diff
@@ -75,7 +75,7 @@ export const getReceivingEmail = (location = getGlobalLocation()) => {
     return process.env.RAZZLE_IRCC_TEST_RECEIVING_ADDRESS
   }
 
-  if (location && location.receivingEmail && !process.env.RAZZLE_STAGE) {
```
timarney: We shouldn't hit this case for RAZZLE STAGE given the sendmail check ya?

```
RAZZLE_IRCC_TEST_RECEIVING_ADDRESS
```

pcraig3: So the sendmail check will fail on startup if 
- we're not running in production (ie, `RAZZLE_STAGE !== 'production'`)
- we haven't set `RAZZLE_IRCC_TEST_RECEIVING_ADDRESS`

This is because when we're not on production we want to always use `RAZZLE_IRCC_TEST_RECEIVING_ADDRESS`. 
Else we'll send emails to the real office.

This check (location.receivingEmail exists && RAZZLE_STAGE is not set) would always fail because RAZZLE_STAGE _is_ set in production.

The logic we want is basically:
- if **not** running in production: return `process.env.RAZZLE_IRCC_TEST_RECEIVING_ADDRESS`
- if **yes** running in production: return `location.receivingEmail`

Issue #398, opened 2018-08-27
New Features


emanelfy:
Adding two new features to the service:
1. "Reschedule my family"
2. "These dates don't work for me"
3. New "Explanation" page 
4. Review page update

--

### Form Page
Order of elements on the page:

- Name
- PFN
- **Schedule family** 
_Should look like this:_
Field should be the same width as the text field used for the reason at the bottom of the page, but 2/3s of the height. The text field only becomes activated and mandatory if the box is checked (otherwise the field is in a disabled grey state).
<img width="654" alt="screen shot 2018-08-27 at 6 27 18 am" src="https://user-images.githubusercontent.com/35348560/44655192-45869e80-a9c2-11e8-8d3c-ebfe2ad2cb01.png">

cc @zakkain - to supply content + error message for if box is checked, but no input in field.

--
- Email
- Radio buttons
- Details field

### Calendar Page
Should be present on page load (and stay regardless of whether calendar selections are made or not).
Left-aligned with the date-box, and aligned with the dates in the last row of the calendar.

**Mobile:** Stacked above the calendar, below the calendar instruction.

Same size and colour as the "cancel" link at the bottom of the screen. 
<img width="933" alt="screen shot 2018-08-27 at 6 20 24 am" src="https://user-images.githubusercontent.com/35348560/44654920-626ea200-a9c1-11e8-9d74-e672c6816051.png">

### Explanation Page
Should only contain a field box that looks like this:
<img width="645" alt="screen shot 2018-08-28 at 10 44 45 am" src="https://user-images.githubusercontent.com/35348560/44730639-68db4780-aaaf-11e8-96f3-2f072e8e2866.png">

@zakkain to supply error message for if a user tries to continue without entering text in the field, and a warning box message to clarify that their request may be reviewed and that they should attend their original appointment, and when staff will respond to users by.

This also changes the nature of the warning messages from that point forward in the flow (review page, etc).

### Review page update
When moving forward in the flow, the dates that would normally appear under the availability section should be replaced by the text inputted in the field on the page prior.
<img width="900" alt="screen shot 2018-08-28 at 10 47 39 am" src="https://user-images.githubusercontent.com/35348560/44730834-d5564680-aaaf-11e8-8fe0-03f5ba8cf88b.png">
Flow progresses normally from here.

emanelfy:
[...]

PR #409, opened 2018-08-29
WIP: These dates dont work

nmakuch:
### Adding the these dates don't work for me link to the Calendar Page (media queries and styling)
 
| Before | After |
|--------|-------|
|    screen shot 2018-08-29 at 15 59 01    |   screen shot 2018-08-29 at 15 58 40    |

### Creating the Explanation Page and adding it to routes.js
- A section of the Registration Page was spliced into a new page named ExplanationPage.js in the repo, route is "/explanation"

screen shot 2018-08-29 at 16 08 53

### TextAreaAdapter Field
- Final form Field has been added to this page using the TextAreaAdapter much like the explanation box on the Registration Page
- 'Continue' button will continue flow to the Review Page

screen shot 2018-08-29 at 16 09 57

### Validation/ErrorMessages for the input field

datesdontwork-valerror 

### Saving Value + Repopulating this value on the Review Page

datesdontwork-saving

### Rerouting flow to skip Calendar if ExplanationPage field data has been saved

datesdontwork-rerouting

### Adding Memory Router to Calendar Tests
- this PR broke some of the calendar tests, the solution was wrapping `<MemoryRouter></MemoryRouter>` around `<CalendarAdapter>`

### Updating .redirect() to accept params
- updated `.redirect(params)` to accept params, this allowed for the rerouting to work when using the Dates Dont Work flow.

### This PR addresses the Dates Don't Work feature in [Issue #398](https://github.com/cds-snc/ircc-rescheduler/issues/398)


web/src/components/Calendar.js
```diff
@@ -412,6 +413,44 @@ const calendarContainerTop = css`
   `)};
 `
 
+const datesLinkBefore = css`
+  margin-top: 19.7rem;
```
pcraig3: Oh no, magic numbers 🙈

web/src/components/Calendar.js
```diff
@@ -412,6 +413,44 @@ const calendarContainerTop = css`
...
+  a {
+    margin-left: 3.7rem;
+
+    ${mediaQuery.lg(css`
+      margin-left: 0;
+      margin: 0;
+    `)};
+  }
+
+  margin-bottom: ${theme.spacing.xxl};
```
pcraig3: Looks like the only difference between `datesLinkBefore` and `datesLinkAfter` is the `margin-top` rule.

Could do this instead to save some lines.

```
const datesLinkBefore = css` 
  margin-top: 19.7rem;
  /* other CSS rules */
`

const datesLinkAfter = css` 
  ${datesLinkBefore};
  margin-top: 5.15rem;
`
```

Since the second `margin-top` comes later in the cascade, it will overwrite the first one.

web/src/pages/ExplanationPage.js
```diff
@@ -0,0 +1,305 @@
...
+        display: block;
+
+        &[id$='-header'] {
+          font-size: ${theme.font.lg};
+          font-weight: 700;
+        }
+      }
+    }
+  }
+`
```
pcraig3: If this CSS is all the same as registration page, we should:
- move into styles.js 
or
- add it to Layout
or 
- `export const contentClass` from RegistrationPage, and then we can use it in here

web/src/pages/ExplanationPage.js
```diff
@@ -0,0 +1,305 @@
...
+      this.props.location.search &&
+      this.props.location.pathname === '/explanation' &&
+      checkURLParams(this.props.location.search, this.fields)
+    ) {
+      errorsNoJS = this.validate(explanation, true)
+    }
+
+    return (
+      <Layout contentClass={contentClass}>
+        <Title path={this.props.match.path} />
```
pcraig3: Should add a page title for this in https://github.com/cds-snc/ircc-rescheduler/blob/master/web/src/components/Title.js

web/src/pages/ExplanationPage.js
```diff
@@ -0,0 +1,305 @@
...
+        </h1>
+        {/*
+          the first checkbox / radio on the page doesn't have its CSS applied correctly
+          so this is a dummy radio button that nobody should ever see
+          it's also outside of the form so it can't be submitted
+          if it is removed, the first radio button in the list of reasons will disappear
+        */}
+        <div style={{ display: 'none' }}>
+          <Radio id="ignore-me" value="ignore-me" />
+        </div>
```
pcraig3: Omg I forgot about this 🤦‍♂️

web/src/pages/ExplanationPage.js
```diff
@@ -0,0 +1,305 @@
...
+    this.validate = ExplanationPage.validate
+    this.fields = ExplanationPage.fields
+    this.redirect = ExplanationPage.redirect
+    this.hasNotValid = this.hasNotValid.bind(this)
+    this.generalErrorMessage = this.generalErrorMessage.bind(this)
+    this.form = null
+  }
+
+  generalErrorMessage() {
+    return <Trans>Some information is missing.</Trans>
```
pcraig3: This message can be made more specific since we know what the problem here is, I would think. 
Two cases are:
- no explanation submitted
- explanation is too long

We could at least mention the explanation in the error message, I would think. 

What do you think, @zakkain?

web/src/pages/RegistrationPage.js
```diff
@@ -108,8 +108,13 @@ class RegistrationPage extends React.Component {
...
 
-  static get redirect() {
-    return '/calendar'
+  static redirect(store = {}) {
+    let { explanation: { explanationPage } = {} } = store
+    return explanationPage ? '/explanation' : '/calendar'
+  }
+
+  get name() {
+    return 'Page'
```
pcraig3: lol, we can remove the `get name()` function.

web/src/components/Calendar.js
```diff
@@ -705,31 +744,40 @@ class Calendar extends Component {
...
+                    removeDayAltText:
+                      i18n !== undefined ? i18n._('Remove day') : 'Remove day',
+                  })}
+                </ul>
+              </div>
+            </div>
+            <div className={value.length ? datesLinkAfter : datesLinkBefore}>
+              <NavLink to="/explanation">
+                <Trans>I&rsquo;m not available for any these days</Trans>
+              </NavLink>
```
pcraig3: Do we want to hide this if 3 dates are selected?

pcraig3: Also, I think you a word.

> I am not available for any these days

should be

> I am not available for any of these days

web/src/pages/ExplanationPage.js
pcraig3: We can actually remove this because we don't have any checkboxes or radios on this page.

web/src/components/Calendar.js
pcraig3: Finally, there's no corresponding link on the noJS page.

web/src/components/Summary.js
```diff
@@ -77,7 +77,8 @@ export const SummaryRow = ({
       <SummaryH2>{summaryHeader}</SummaryH2>
       <SummaryBody>{summaryBody}</SummaryBody>
     </SummaryHeader>
-    {summaryLink === '/register#explanation-label' ? (
+    {summaryLink === '/register#explanation-label' ||
+    '/register#availabilityExplanation-label' ? (
```
pcraig3: This should be `/explanation#explanationPage-label`

web/src/components/Summary.js
```diff
@@ -141,12 +143,22 @@ const Summary = ({
...
+        summaryHeader={<Trans>Availability</Trans>}
+        summaryBody={<SelectedDayList selectedDays={selectedDays} />}
+        summaryLink={'/calendar#calendar-header'}
+        summaryLabel={i18n && `${i18n._('Change')} ${i18n._('Availability')}`}
+      />
+    ) : (
+      <SummaryRow
+        summaryHeader={<Trans>Availabilty</Trans>}
+        summaryBody={availabilityExplanation}
+        summaryLink={'/register#availabilityExplanation-label'}
```
pcraig3: This link should be `/explanation#explanationPage-label` (also mentioned above ☝️)

web/src/pages/ExplanationPage.js
pcraig3: Doing the same thing as the `h1` makes the most sense.

web/src/pages/ReviewPage.js
```diff
@@ -93,13 +98,21 @@ class ReviewPage extends React.Component {
...
+              <Trans>
+                You should plan to attend your existing appointment until we
+                contact you. This may take 1 week.
+              </Trans>
+            ) : (
+              <Trans>
+                Sending this request will cancel your current appointment.
+                <strong> Do not attend your old appointment</strong> after you
+                send this request.
+              </Trans>
```
pcraig3: I don't think `<strong>` tags inside of `<Trans>` tags will work.

Probably what we have to do here is

```
<React.Fragment>
  <Trans>Sending this request will cancel your current appointment.</Trans>
  <strong>
    <Trans>Do not attend your old appointment</Trans>
  </strong>
  <Trans>after you send this request.</Trans>
</React.Fragment>
```

pcraig3:
This PR is on hold until some design changes are made. 

We want to get in #404 before this one goes in.

web/src/pages/ExplanationPage.js
zakkain: Error message for no explanation here: https://github.com/cds-snc/ircc-rescheduler/issues/398#issuecomment-417019719

for "explanation is too long" 👉 `Please be more concise, your answer is too long.`

PR #411, opened 2018-08-29
500 error page

timarney:
Adds 500 error page

Adds a generic page where users can be redirected if we hit a server error.

Use cases:
no location exist
receiving address not set


web/src/components/Footer.js
```diff
@@ -106,11 +106,12 @@ const Footer = ({ topBarBackground, i18n, match = { url: '' } }) => (
     {topBarBackground ? <TopBar background={topBarBackground} /> : ''}
     <footer id="footer" className={footer}>
       <div className={bottomLinks}>
-        {match.url !== '/not-found' && (
-          <a href={`mailto:${getEmail()}`}>
-            <Trans>Contact</Trans>
-          </a>
-        )}
+        {match.url !== '/not-found' &&
+          match.url !== '/500' && (
```
pcraig3: What do you think of passing something through Layout that gets passed to Footer instead of doing these checks? 

Like
```
<Layout contact={false}>
```
or something?

timarney: That sounds good and would work for the 500 page but the 404 page isn't a fresh page used for /not-found + /really-missing-page how would it work in that case?

pcraig3: Haha! [404 page is already checking for `/not-found`](https://github.com/cds-snc/ircc-rescheduler/blob/master/web/src/pages/FourOhFourPage.js#L34)

So could be like `<Layout contact={!(url === 'not-found')} or something.

web/src/components/Footer.js
```diff
@@ -101,12 +100,12 @@ const TopBar = styled.hr(
   props => ({ background: props.background }),
 )
 
-const Footer = ({ topBarBackground, i18n, match = { url: '' } }) => (
+const Footer = ({ contact, topBarBackground, i18n }) => (
```
pcraig3: We should probably default `contact` to true in here. 

The tests are failing because it's getting `undefined` and the proptypes say it should be a boolean.

web/src/components/Layout.js
```diff
@@ -85,6 +85,7 @@ class Layout extends React.Component {
   }
 
   render() {
+    const { contact } = this.props
```
pcraig3: Could do `const { contact = true } = this.props` in here (ie, set contact to true if it doesn't exist in props), and then just pass it in to Footer directly.

web/src/pages/FiveHundredPage.js
```diff
@@ -0,0 +1,31 @@
...
+  p {
+    padding-bottom: ${theme.spacing.lg};
+  }
+`
+
+class FiveHundredPage extends React.Component {
+  render() {
+    return (
+      <Layout contentClass={contentClass} contact={false}>
+        <Title path={this.props.match.path} />
```
pcraig3: If we're adding a Title, we should also add a page <title> in the component itself: https://github.com/cds-snc/ircc-rescheduler/blob/master/web/src/components/Title.js#L39

web/src/server.js
```diff
@@ -101,7 +101,7 @@ const _ensureBody = (req, res, next, cb) => {
   try {
     cb()
   } catch (e) {
-    Raven.captureException(e)
+    //Raven.captureException(e)
```
pcraig3: We can also uncomment this now, right?

web/src/components/Footer.js
pcraig3: `Footer.defaultProps` → 👍

PR #413, opened 2018-08-29
Make redirect domain check more inclusive

pcraig3:
Previously, we would redirect you on staging, but not on production.

Since we never deployed to production, this has never come up, but let's fix it now rather than wait and see 😊


pcraig3:
Closing in favour of #414

PR #414, opened 2018-08-29
Enforce subdomain

timarney:
When a user is visiting the website using the staging or live url + the base sub-domain we need to redirect to a proper location sub-domain.

Later we will 404 the use of the base sub-domain but this case needs to be handled for existing links.

**Testing manually :**
Use: http://rescheduler-dev.vcap.me:3004 & http://rescheduler.vcap.me:3004

### DIFF
--- a/web/src/server.js
+++ b/web/src/server.js
@@ -10,7 +10,6 @@ import { renderStylesToString } from 'emotion-server'
 import bodyParser from 'body-parser'
 import { CalendarFields, RegistrationFields } from './validation'
 import Validator from 'validatorjs'
-// import cache from 'memory-cache'
 import {
   getMailer,
   getEmailParms,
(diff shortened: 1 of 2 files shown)


web/src/server.js
```diff
@@ -73,20 +72,47 @@ const captureMessage = (title = '', validate) => {
...
   }
 
+  /* 
+  If no sub-domain is found
+  force redirect to vancouver sub-domain on staging and prod
+  note: this is temporary to handle existing vancouver traffic / links
+  */
+
+  if (forceRedirect(req)) {
+    return res.redirect(`https://vancouver.${process.env.RAZZLE_SITE_URL}`)
```
pcraig3: This here doesn't respect the protocol. Redirects you locally to `https://vancouver.vcap.me:3004/`

PR #421, opened 2018-09-05
Clean up server.js

timarney:
Server.js has become larger over time taking on a bunch of functions that would be better split out into utilities.  This streamlines server.js back to basics making it easier to see what's going on.

## Notes:
- Moves the locations whitelist into the locations directory - keeping all the location functionality in a single place

- Moves sending email validation & logic into the email directory

- Moves Raven and anything that was a standalone function into a utilities file.


pcraig3:
oh no!

image

PR #423, opened 2018-09-05
Limit emails sent by tests

timarney:
The tests are firing multiple emails every time there is a PR.  This limits the emails sent per run down to a single test that does a full run with validation errors.

Ideally with less emails being sent if something is off we'll notice sooner.

@pcraig3 ultimately it would be nice to find something that would verify receipt to an inbox.  Looking to at least cut down on the noise for now.


pcraig3:
😄 

```
{
  cy.visit('/confirmation')
}

cy.url().should('contain', '/confirmation')
```

timarney:
Found a couple more using Full Run.

PR #427, opened 2018-09-07
Update no js pattern for yarn test

timarney:
Running yarn test locally is failing the NoJS tests. We shouldn't be including the NoJS tests for that test run.  Use a wildcard when looking for the NoJs tests.

Thinking for CI we should split them out into separate runs.  Running nojs:tests  as a separate step.


timarney:
Handle via - https://github.com/cds-snc/ircc-rescheduler/pull/428

PR #428, opened 2018-09-07
Setup feature flags

timarney:
What's a feature flag?

A feature flag provides an alternative to maintaining multiple feature branches, such that a feature can exist before it's ready for release. Features can be enable or disabled the at run time.  I.e. we can hide a feature until we're ready to deploy it to all users.

For this setup a flags env variable has been added to allow toggling features

```
RAZZLE_FLAGS="[{"name":"nextButton","isActive":true}]"
```

We can now use the flags using the renderOn and renderOff properties

```
<Flags
    authorizedFlags={['nextButton']}
    renderOn={() => (
     <Button disabled={submitting}>
           <Trans>Next</Trans>
     </Button>
   )}
  renderOff={() => (
   <Button disabled={submitting}>
    <Trans>Review request</Trans>
     </Button>
       )}
   />
```


web/src/pages/CalendarPage.js
```diff
@@ -401,9 +402,19 @@ class CalendarPage extends Component {
...
+                        authorizedFlags={['nextButton']}
+                        renderOn={() => (
+                          <Button disabled={submitting}>
+                            <Trans>Next</Trans>
+                          </Button>
+                        )}
+                        renderOff={() => (
+                          <Button disabled={submitting}>
+                            <Trans>Review request</Trans>
+                          </Button>
```
pcraig3: For some reason (and I'm not sure why), this doesn't work on the noJS version of the page.

When you do a hard refresh on the calendar page, you will see the button change back and forth if you have come from the landing page. Even though when we're printing all the values on the server, it's always printing out the right flags.

Not sure how the `Flags` component is deciding which version to show, but it seems to work in some cases but not others.

flags

web/src/utils/flags.js
```diff
@@ -0,0 +1,18 @@
+export const parseFlags = () => {
+  let flags = []
+  try {
+    if (process.env.RAZZLE_FLAGS) {
+      flags = JSON.parse(process.env.RAZZLE_FLAGS)
+      // eslint-disable-next-line no-console
+      console.log(
```
pcraig3: Do we want to keep this around? As it stands, it will be printed in all our users' browser consoles.

timarney: I guess that depends on the plan - will we have flags set for production? 

Will remove as it's not really needed.

web/src/pages/CalendarPage.js
timarney: Maybe takes x ms to run this filter

https://github.com/romaindso/react-feature-flags/blob/master/src/index.js#L45

timarney: Maybe add a windowExists check and avoid this ... given the features will be for testing.

pcraig3: Ah right, if it defaults to not showing them, we don't really mind because in prod we never want them there.

Seems like a reasonable justification to me.

web/src/utils/flags.js
```diff
@@ -0,0 +1,12 @@
...
+  let flags = []
+  try {
+    if (process.env.RAZZLE_FLAGS) {
+      flags = JSON.parse(process.env.RAZZLE_FLAGS)
+    }
+  } catch (e) {
+    // do nothing
+  }
+
+  return flags
```
pcraig3: If this returns an empty array, the `Flags` just won't be called eh?

timarney: Won't match on a flag

PR #429, opened 2018-09-07
Turn noJS end-to-end tests back on in CircleCi

pcraig3:
## waitFor after clicking page navigation

We were seeing a lot of intermittent failures running `yarn test:full` with the puppeteer tests giving us an `Execution context was destroyed` error.

Seems like that error message is trying to tell us that [the page is being navigated away from in a way that puppeteer doesn't like](https://github.com/GoogleChrome/puppeteer/issues/1591#issuecomment-355245777).  [Someone suggested using page.waitFor](https://github.com/GoogleChrome/puppeteer/issues/1591#issuecomment-363384982) because that solved their issue. Seems to work for our case.

Created a new function `clickAndWait` that will call `page.click` and then `page.waitFor`. Running this a bunch of times locally and it seems to pass reliably now.

## puppetter-config.js

Lots of our setup (user fixtures + debugging methods, etc) is duplicated across files.

Created a new file to hold all of the shared code we use in our puppeteer tests.

Keeping it in a `__tests__` directory meant I had to change the jest config to not try to run _this_ file as a test. If we think that's a bad idea, we can put this somewhere else maybe.


pcraig3:
Thanks! Ran this 4 or 5 times on CI and never saw it fail after bumping the `waitFor` delay up to `500`ms.

PR #430, opened 2018-09-07
NoJS validation error when users have submitted family members but have not checked the "I need to reschedule my family too"

pcraig3:
## JS validation

In the JS flow, we disable the family member textarea, which means you can't accidentally enter family members names into it without clicking the checkbox. It's also visually de-emphasized, so we're pretty confident that if someone has something left in the textarea but they uncheck the box, they don't want to rebook their family members. So in the JS case, we erase their `familyOption` value if `familyCheck` is missing.

|                      | textarea empty | textarea has a value |
|----------------------|----------------------|------------------|
| checkbox not checked □      | 👍                    |       👍          |
| checkbox checked    ✓ | 🛑                   | 👍                |

## NoJS validation

In the NoJS flow, we have to keep the textarea enabled, so people are able to enter family members in, although they won't be saved unless the checkbox is clicked as well.

Instead of silently deleting user data if they haven't checked the checkbox, we want to validate them on the noJS flow if they've entered family members but not clicked the checkbox.

|                      | textarea empty | textarea has a value |
|----------------------|----------------------|------------------|
| checkbox not checked □      | 👍                    |   🛑          |
| checkbox checked    ✓ | 🛑                   | 👍                |

### Gifs

| JS validation | NoJS validation |
|---------------|-----------------|
|  js-val   |   no-js-val    |


pcraig3:
Note that I am not adding translations to this yet as
- it would mess up the diffs
- we want to have @zakkain look at the english
- we want a good french translation

web/src/pages/RegistrationPage.js
```diff
@@ -87,6 +91,8 @@ const labelNames = id => {
       return <Trans>Why are you rescheduling?</Trans>
     case 'explanation':
       return <Trans>Describe why you can’t attend your appointment</Trans>
+    case 'familyCheck':
+      return <Trans>Please confirm you are resheduling family members</Trans>
```
pcraig3: FYI: "resheduling" is spelled wrongly.

PR #432, opened 2018-09-10
Update explain page

timarney:
Updates explain page per https://github.com/cds-snc/ircc-rescheduler/issues/398#issuecomment-418880887


web/src/pages/ExplanationPage.js
```diff
@@ -156,19 +163,24 @@ class ExplanationPage extends React.Component {
     return (
       <Layout contentClass={explanationContentClass}>
         <Title path={this.props.match.path} />
+        <NavLink className="chevron-link nav-link-top" to="/calendar">
```
pcraig3: What we've done so far is to ignore the calendar page if an `explanation` exists.

This is the right link back if someone has come from the calendar page and wants to go back, but if they've submitted an explanation and are coming back from the "/review" page, we should link you back to the "/register" page instead.

timarney: Right - best approach would be to check against the history object?

web/src/pages/ExplanationPage.js
```diff
@@ -156,19 +166,29 @@ class ExplanationPage extends React.Component {
     return (
       <Layout contentClass={explanationContentClass}>
         <Title path={this.props.match.path} />
+        <a
+          className="chevron-link nav-link-top back"
+          onClick={() => {
+            this.props.history.goBack()
+          }}
+        >
```
pcraig3: A couple problems with this:
- breaks our SPA flow because it's a real link
- doesn't work without JS (because there is no `href`) 

What we did in other pages was something like
```
<Navlink to={ explanationPage ? "/register" : "/calendar" }  > ... </NavLink>
```

If you can get that as a string out of `props.history`, then your way is probably better though.

timarney:
props.history doesn't allow reading the previous location.

Thinking of adding a ?reviewed=true  which I could check to figure out what page to go back to.

Thoughts?

web/src/pages/ExplanationPage.js
```diff
@@ -227,8 +244,8 @@ class ExplanationPage extends React.Component {
                       />
                       <span id="explanationPage-details">
                         <Trans>
-                          Provide enough detail so that staff can understand
-                          your situation.
+                          If you haven’t already, please provide a reason for
+                          your unavailability
```
pcraig3: This should end with a period. On the register page, the bold headings don't end in periods, but the subheader descriptions do.

web/locale/fr/messages.json
```diff
@@ -21,6 +21,15 @@
...
   },
+  "Apply for an appointment extension": {
+    "translation": "Demander une extension de rendez-vous",
+    "origin": [
+      [
+        "src/components/Title.js",
+        27
+      ]
+    ]
+  },
```
pcraig3: I don't know if you want to change it in this PR, but [the link from the calendar page](https://github.com/cds-snc/ircc-rescheduler/blob/master/web/src/components/Calendar.js#L771) hasn't been updated with @emanelfy's new copy.

web/src/validation.js
```diff
@@ -87,9 +87,7 @@ errorMessages.explanationMaxErrorMessage = (
 )
```
pcraig3: Something else I noticed while clicking around is that `Sorry, there‘s a limit of 150 words` is using the wrong smart quote. It's not so obvious on github, but super obvious in the app.

image


Gotta flip it around. Should be `’`.

web/src/components/Title.js
```diff
@@ -24,7 +24,7 @@ const Title = ({ i18n, path }) => {
       title = `${i18n._('Select 3 days you’re available')} ${divider} ${title}`
       break
     case '/explanation':
-      title = `${i18n._('Delay your appointment')} ${divider} ${title}`
+      title = `${i18n._('Apply for an appointment extension')} ${divider} ${title}`
```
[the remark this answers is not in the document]
zakkain: Yep, 💯

web/src/pages/ReviewPage.js
```diff
@@ -131,6 +131,7 @@ class ReviewPage extends React.Component {
             explanation={explanation}
             reason={reason}
             selectedDays={selectedDays}
+            availabilityExplanation={explanationPage}
```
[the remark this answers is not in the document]
zakkain: I think I need to refresh my memory a bit , but `Availability` still works well, because this would be the header for the big blurb they give, right?

PR #439, opened 2018-09-11
Family calendar

nmakuch:
### 1) Updating the page header on the Calendar Page for people going through the Family Application process:
- The message in the Calendar Header is now conditional. It will switch based on whether the applicant is filling out a family application or solo application.

| Solo Application | Family Application |
|------------------|--------------------|
|  screen shot 2018-09-11 at 11 40 44 |   screen shot 2018-09-11 at 11 39 49  |


pcraig3:
Added some cypress tests that check for the new header text and refactored the tests a bit while doing it. Should be good to merge now.

PR #441, opened 2018-09-12
Update email templates to handle availability explanation

timarney:
With the addition of https://github.com/cds-snc/ircc-rescheduler/pull/438 the email templates need to handle either selected days or the availability explanation.  This updates handles the new scenarios and is backwards compatible with the existing setup.

## Testing
To test locally run update the cypress user fixture to your email address and run "validation-error.spec" (Full Run-through including validation errors).  To test with an availability explanation add an explanation manually i.e. 

```
options.formValues.availabilityExplanation = "Hey this works!"

if (
    typeof options.formValues.availabilityExplanation !== 'undefined' &&
    options.formValues.availabilityExplanation !== ''
  ) {
    renderAvailabilityExplanation(options)
  } else {
    renderSelectedDays(options)
  }

...
``` 

Sample user email:
screen shot 2018-09-12 at 1 25 48 pm

### DIFF
--- a/web/cypress/fixtures/user.json
+++ b/web/cypress/fixtures/user.json
@@ -1,7 +1,7 @@
 
 {
 "fullName": "Leanne Graham",
-"email": "cdc@example.com",
+"email": "cds@example.com",
 "paperFileNumber": "123456",
 "reason": "workOrSchool",
 "explanation": "On business trip!",
--- a/web/src/email/email-template.js
+++ b/web/src/email/email-template.js
@@ -100,11 +100,11 @@ const renderMarkup = async options => {
   ])
 }
 
-export const buildParams = async options => {
-  const { selectedDays } = options.formValues
-
-  options.formValues.respondByDate = respondByDate(selectedDays, 'en')
-  options.formValues.respondByDateFR = respondByDate(selectedDays, 'fr')
+const renderSelectedDays = options => {
+  const selectedDays = options.formValues.selectedDays
+  options.formValues.datesLabel = 'These are the days you selected:'
+  options.formValues.datesLabelFR =
+    'Voici les journées que vous avez sélectionnées:'
 
   // add line breaks to dates available
   options.formValues.datesHtml = datesMarkup(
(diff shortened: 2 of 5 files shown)


web/cypress/fixtures/user.json
```diff
@@ -1,7 +1,7 @@
 
 {
 "fullName": "Leanne Graham",
-"email": "cdc@example.com",
+"email": "tim@line37.com",
```
pcraig3: Does this mean that you will get these emails from now on on every PR?

web/src/email/email-template.js
```diff
@@ -129,6 +129,45 @@ export const buildParams = async options => {
...
+}
+
+const renderAvailabilityExplanation = options => {
+  const availability = options.formValues.availabilityExplanation
+  options.formValues.datesLabel = 'When are you unavailable?'
+  options.formValues.datesLabelFR = 'Quand êtes-vous indisponible?'
+  options.formValues.datesHtml = availability
+  options.formValues.datesHtmlFR = availability
+  options.formValues.datesPlain = availability
+  options.formValues.datesPlainFR = availability
```
pcraig3: Right, we don't need to translate these at all because we are just putting their literal text in the email.  👍

web/src/email/email-template.js
```diff
@@ -129,6 +129,45 @@ export const buildParams = async options => {
...
+
+  options.formValues.whatHappensNext =
+    '<p>You should plan to attend your existing appointment until we contact you. This may take 1 week.</p>'
+
+    options.formValues.whatHappensNextFR =
+    '<p>Vous devriez prévoir assister à votre rendez-vous actuel jusqu’à ce que nous communiquions avec vous. Cela pourrait prendre jusqu’à une semaine.</p>'
```
pcraig3: If either of these lines changes in the app in the future, we should make a mental note to remember to change them in the email. Not sure if we would have a better was of doing that.

Thoughts: 
- we could try directly including the messages.json string in here
- (that would mean that the key would be the literal string though)

Not sure if that would be more brittle for little added benefit.

web/src/email/email-template.js
```diff
@@ -129,6 +129,45 @@ export const buildParams = async options => {
...
+  const { selectedDays } = options.formValues
+
+  options.formValues.respondByDate = respondByDate(selectedDays, 'en')
+  options.formValues.respondByDateFR = respondByDate(selectedDays, 'fr')
+
+  // render selected dates or unavailability
+  if (
+    typeof options.formValues.availabilityExplanation !== 'undefined' &&
+    options.formValues.availabilityExplanation !== ''
+  ) {
```
pcraig3: Can you think of an edge case where a truthy check wouldn't work?

Ie, since an empty string is falsey, we could just have the condition be: `if(options.formValues.availabilityExplanation)`

***

And, to be even more terse, you could also declare the variable on line 156, like `const {selectedDays, availabilityExplanation} = options.formValues`

Anyway, just suggestions if you wanted to take them.

web/src/email/email-template.js
timarney: Could maybe add an HTML comment in the template to note this.  

I don't think pulling from messages.json will work as we might not always have an exact match for the strings.

web/src/email/email-template.js
timarney: I'll try again but using just options.formValues.availabilityExplanation seemed to throw an error (might have been something else).

web/cypress/fixtures/user.json
```diff
@@ -1,7 +1,7 @@
 
 {
 "fullName": "Leanne Graham",
-"email": "cdc@example.com",
+"email": "cds@exemple.com",
```
pcraig3: exemple -> example

Issue #447, opened 2018-09-13
Header: include sentence around "language selection" between BETA and "francais"


timarney:
Users expect full list of available languages (not just "Francais")

Implication:
The first thing the user hears is "beta francais" and they don't know if they need to take an action to select the right language or if english is an option.

Visually the language selection will remain the same.
 
@zak should this read -> language selection Français ?

timarney:
@zak any thoughts on translation 

## Viewing EN
language selection: Français

## Viewing FR
language selection: English or should this be -> sélection de la langue: English

PR #453, opened 2018-09-14
WIP - Announce language toggle

timarney:
To confirm how the text will read on EN vs FR
https://github.com/cds-snc/ircc-rescheduler/issues/447

### DIFF
--- a/web/src/components/LanguageSwitcher.js
+++ b/web/src/components/LanguageSwitcher.js
@@ -94,8 +94,8 @@ class LanguageSwitcher extends React.Component {
             className={visuallyhiddenMobile}
             aria-label={
               this.getNewLanguage() === 'fr'
-                ? 'language selection: Français'
-                : 'sélection de la langue:  English'
+                ? 'Language selection: French'
+                : 'Sélection de la langue: anglais'
             }
           >
             {this.getNewLanguage() === 'fr' ? 'Français' : 'English'}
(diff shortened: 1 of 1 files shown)


emanelfy:
On english should say "french"
On french should say "anglais"

web/src/components/LanguageSwitcher.js
```diff
@@ -90,7 +90,14 @@ class LanguageSwitcher extends React.Component {
             logEvent('Navigation', 'Toggle Language', lang)
           }}
         >
-          <span className={visuallyhiddenMobile}>
+          <span
+            className={visuallyhiddenMobile}
+            aria-label={
+              this.getNewLanguage() === 'fr'
+                ? 'language selection: Français'
+                : 'sélection de la langue:  English'
```
pcraig3: After talking to @emanelfy, the two cases we want are:
- If the page is in English: `Language selection: French`
- If the page is in French: `Sélection de la langue: anglais`

Since this is the first content that shows up if you tab through the page, this is potentially the first thing that someone with a screenreader will get. If the rest of the page is in English but the first thing they read is in French (ie, "Francais"), then they may not know they are on an English-language page. 

So this way, it matches the primary language of the rest of the page. 

Let's change to that and then re-assess.

timarney:
@emanelfy 
So the full announced version would say

**Viewing english**
language selection: french

**Viewing french**
sélection de la langue:  anglais

Correct?

PR #460, opened 2018-09-18
Prepend a comma and space to familyOption string in emails

### LINKED ISSUE #443: Formatting and email treatment of plaintext family members field
https://github.com/cds-snc/ircc-rescheduler/blob/a13d81a819f2cf503870a13349bb2f352a784b58/web/email_templates/applicant-rich.html#L43-L44

- Discussed the issues around making opinionated formatting decisions with Paul. Some solutions I recommend:

image

1. Change `Example:` to `For example:`
2. Put the quotation marks around `Full Name, Full Name, Full Name` but exclude `For example:` The "example" isn't quoted, only the names are.
3. Same changes apply for French.

For the email, I think we _have_ to remain un-opinionated about text in that field, and keep that formatting in the email. Sure, some people might get really long, messy emails, but that might actually look "correct" because that formatting is how they input their text. This is something to observe, IMO, and if folks are putting descriptive info in that field, we can modify the description/instruction copy to either accommodate that, or specifically recommend against it.

pcraig3:
Previously, if your name was "Paul Craig" and your familyOption string was "Julia Craig, Kevin Craig" we would have:

- `Paul Craig Julia Craig, Kevin Craig`

Now we will have

- `Paul Craig, Julia Craig, Kevin Craig`

This doesn't solve every case, but I would assume it's an improvement in the vast majority of cases.

*Note: Since `options.formValues` is an object we keep in memory between function calls, we have to check that a comma and a space have not already been prepended.*

This closes #443

### DIFF
--- a/web/src/email/email-template.js
+++ b/web/src/email/email-template.js
@@ -153,7 +153,11 @@ const renderAvailabilityExplanation = options => {
 }
 
 export const buildParams = async options => {
-  const { selectedDays, availabilityExplanation } = options.formValues
+  const {
+    selectedDays,
+    availabilityExplanation,
+    familyOption,
+  } = options.formValues
 
   options.formValues.respondByDate = respondByDate(selectedDays, 'en')
   options.formValues.respondByDateFR = respondByDate(selectedDays, 'fr')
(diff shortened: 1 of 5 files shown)


web/src/email/email-template.js
```diff
@@ -165,6 +169,12 @@ export const buildParams = async options => {
     renderSelectedDays(options)
   }
 
+  // if familyOption exists and does not start with ', ', prepend a comma and a space
+  options.formValues.familyOption =
+    familyOption && !familyOption.startsWith(', ')
```
timarney: Thinking we should just check for ',' here thoughts?

pcraig3: Sure, easy enough.

PR #461, opened 2018-09-19
Custom 'not-allowed' cursor + Green Hover (Unavailability Flow)

nmakuch:
### 1) Custom 'not-allowed' cursor:
- Updated project to have a custom cursor when in Chrome or Firefox (we felt the default 'not-allowed' cursor on these browsers  was not providing enough contrast for users to notice it)

| Before | After |
|--------|-------|
|   notallowed-before     |  notallowed-after     |

### 2) Green Hover on selectable dates:
- The grey that was being used for the selectable dates hover was too close to the colour being used for the disabled dates (on low quality screens and projectors, these colours look almost identical). We have switched this to be a light shade of green (40% less opacity than `theme.colour.green`)

| Before | After |
|--------|-------|
|    hover-before    |   hover-after    |

#### Color meets WCAG AA + AAA guidelines for contrast
screen shot 2018-09-19 at 12 10 51


pcraig3:
Gonna merge this to unblock later PRs.

PR #465, opened 2018-09-20
Add full date 'x' icon (remove selected day)

timarney:
@pcraig3 thoughts on best way to test this?  Should we have it on the button and the image?

### DIFF
--- a/web/src/components/MobileCancel.js
+++ b/web/src/components/MobileCancel.js
@@ -2,14 +2,19 @@ import React from 'react'
 import PropTypes from 'prop-types'
 import { theme } from '../styles'
 
-const MobileCancel = ({ circleColour = theme.colour.blackLight }) => {
+const MobileCancel = ({
+  circleColour = theme.colour.blackLight,
+  label = '',
+}) => {
   return (
     <svg
       xmlns="http://www.w3.org/2000/svg"
       data-name="Layer 1"
       viewBox="0 0 21.62 21.62"
-      alt=""
+      role="img"
+      aria-labelledby="titleId"
     >
+      <title id="titleId">{label}</title>
       <g data-name="Layer 2">
         <g data-name="Layer 1-2">
           <circle cx="10.81" cy="10.81" r="10.81" fill={circleColour} />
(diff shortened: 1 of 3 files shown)


web/src/components/__tests__/Calendar.test.js
```diff
@@ -453,7 +453,6 @@ describe('renderDayBoxes', () => {
 
     let imgs = button.find('svg')
     expect(imgs.length).toBe(1)
-    expect(imgs.at(0).props().alt).toEqual('')
```
pcraig3: Does `imgs.at(0).props().alt` return anything (ie, that we can test against), or do we not know because it's some weird i18n object?

web/src/components/MobileCancel.js
```diff
@@ -2,13 +2,16 @@ import React from 'react'
...
+  circleColour = theme.colour.blackLight,
+  label = '',
+}) => {
   return (
     <svg
       xmlns="http://www.w3.org/2000/svg"
       data-name="Layer 1"
       viewBox="0 0 21.62 21.62"
-      alt=""
+      alt={label}
```
pcraig3: So I've read a couple articles about this before and there are two recommended approaches I've seen for `svg`s specifically.

1. Add `<title>` tag and `aria-label`.
Code example:
```html
<svg role="img" aria-label="Title text">
  <title>Title Text</title>
  <g data-name="Layer 2"> ... </g>
</svg>
```
Source: https://stackoverflow.com/a/4756461/9728185

1. Add `<title>` tag with id and `aria-labelledby`.
Code example:
```html
<svg role="img" aria-labelledby="titleId">
  <title id="titleId">Title Text</title>
  <g data-name="Layer 2"> ... </g>
</svg>
```
Source: 
 - https://css-tricks.com/accessible-svgs/#article-header-id-6
 - https://stackoverflow.com/a/30717940/9728185

As to your other question, to test this we can only really try with voiceover. If you pull the svg outside of the button and add a tabindex, you can focus with your keyboard and listen to what voiceover does.
If it's the same both ways, then either one is good.

web/src/components/__tests__/Calendar.test.js
timarney: Updated the tests :)

PR #466, opened 2018-09-20
Explicitly set focus on every page

timarney:
Adds a FocusedH1 component that will automatically be given page focus when it loads.

@pcraig3 should we add this to pages line Cancel as well?


web/src/components/Layout.js
```diff
@@ -101,7 +105,7 @@ class Layout extends React.Component {
             <FederalBanner />
             <PageHeader>{this.props.header}</PageHeader>
           </div>
-          <main role="main">
+          <main>
```
pcraig3: This is something that's flagged by most HTML validation tools, but it's something that's useful for older internet explorers, [as mentioned earlier](https://github.com/cds-snc/ircc-rescheduler/issues/78#issuecomment-385426486).

Here's the link: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/main#Browser_compatibility

web/src/pages/CalendarPage.js
```diff
@@ -326,7 +305,7 @@ class CalendarPage extends Component {
     const { month } = this.state
 
     return (
-      <Layout>
+      <Layout focusContent={false}>
```
pcraig3: Don't need this anymore ✂️

web/src/components/FocusedH1.js
```diff
@@ -0,0 +1,54 @@
...
+import { css } from 'react-emotion'
+import { theme } from '../styles'
+
+const h1Style = css`
+  font-size: ${theme.font.xl};
+  font-weight: 400;
+  margin-bottom: ${theme.spacing.xl};
+  margin-top: 0;
+  outline: 0;
+`
```
pcraig3: Haven't checked myself, but [the font-size is larger in `styles.js`](https://github.com/cds-snc/ircc-rescheduler/blob/master/web/src/styles.js#L139). Does this CSS change any of our layouts?

timarney: Nice catch.

Pulled the sizing over from the Cal Page:
```
const CalendarHeader = styled(H1)`
  font-size: ${theme.font.xl};
  ${headerStyles};
`
```

I'll adjust to use theme.font.xxl and then mod for just the Cal page again.

web/src/pages/CalendarPage.js
```diff
@@ -81,6 +69,9 @@ const CalReminder = styled(Reminder)`
 const fullWidth = css`
   width: 100% !important;
 `
+const H1 = css`
+  font-size: ${theme.font.xl};
+`
```
pcraig3: The convention we've been using is to only use capital letter names for Components proper (ie, `const H1 = styled.h1`) but smaller case for just CSS. So maybe `calH1` would work better (or maybe you could directly add it to `style` and then extend the params `FocusedH1` takes).

web/src/components/CalendarH1.js
```diff
@@ -0,0 +1,31 @@
...
+        {getStartMonthName(new Date(), locale)} <Trans>and</Trans>{' '}
+        {getEndMonthName(new Date(), locale)}.
+      </React.Fragment>
+    )
+  }
+}
+
+CalendarH1.propTypes = {
+  familyOption: PropTypes.string,
+  locale: PropTypes.string,
```
pcraig3: Let's make this `isRequired` since we always want it to exist.

PR #468, opened 2018-09-21
Set global 'extra' context for raven

pcraig3:
For our manually triggered Raven errors (ie, `Raven.captureException` or `Raven.captureMessage`), Raven isn't capturing all the contextual data automatically, the way it does when it captures a `throw new Error` event.

What this means is that we have a lot of errors showing up in sentry because our tests trigger them, but we're not able to at-a-glance know when they're coming from our production application.

By setting a global context.extra, all exceptions and messages captured will have the following useful context:
- `tags`
  - the url host name
  - the subdomain
  - the url path
  - the RAZZLE_STAGE env var
- `extra`
  - the release string

**Note: since the context is only getting set on server reloads, the req.path will not be up to date unless the error happens on the first page. I think this is fine, but just so we know.**

## screenshot (new tags highlighted in red)

sentry io_canadian-digital-service_ircc-rescheduler_issues_675614634__query is_unresolved

### DIFF
--- a/web/src/server.js
+++ b/web/src/server.js
@@ -12,6 +12,7 @@ import {
   getPrimarySubdomain,
   ensureLocation,
   ensureReceivingEmail,
+  setRavenContext,
 } from './utils/serverUtils'
 import { handleSubmitEmail } from './email/handleSubmitEmail'
 import gitHash from './utils/gitHash'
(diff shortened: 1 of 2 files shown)


pcraig3:
🎉

PR #472, opened 2018-09-21
Give focus to selected day box

timarney:
When coming back from the review page this will give focus to the selected day box to make it eaier to remove days before moving back to the calendar.

### DIFF
--- a/web/src/components/Calendar.js
+++ b/web/src/components/Calendar.js
@@ -751,9 +751,6 @@ class Calendar extends Component {
             tabIndex="-1"
             className={removeDateMessage}
             id="removeDateMessage"
-            ref={removeDateContainer => {
-              this.removeDateContainer = removeDateContainer
-            }}
           >
             <h2>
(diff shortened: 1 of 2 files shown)


pcraig3:
Couple things:
- Not sure if we do want an outline -- looks worse with it, imo
- Not sure if we want to update the hash (#calendar-header) to somewhere else (#removeDateMessage) for example

Can ask @emanelfy on Monday

timarney:
> * Not sure if we want to update the hash (#calendar-header) to somewhere else (#removeDateMessage) for example

Updated the return link to use /calendar#selected-day-box.  The other hash didn't work well for the mobile view.

emanelfy:
**Outstanding change on the calendar page:**
When a selection is made on the calendar, the active colour should be green (to match the same colour as the button).

Additionally, on the above point -- no outline required :)

timarney:
Making the cal colour updates here https://github.com/cds-snc/ircc-rescheduler/pull/479

PR #474, opened 2018-09-24
pass env variables into yarn sentry_map command

dsamojlenko:


entrypoint.sh
```diff
@@ -1,7 +1,5 @@
 #!/bin/sh
 
 COMMIT_HASH=`cat /web/VERSION`
```
pcraig3: This could be more easily written as `🐱 🕸/VERSION`

PR #475, opened 2018-09-24
Cal page flow links for unavailability

timarney:
Update links for the unavailability flow to ensure users can re-reach the calendar after filling in the explanation page.


pcraig3:
Made a couple of changes:

1. `redirect` used to be a getter when it was just a static string. Since it's no longer static, we can make it a getter again.

2. Added `explanationPage` to the list of valid `CalendarPage` fields 

<details>
<summary>👈 click here for more details</summary>  

Initially ran into a problem where the `explanationPage` value was not being reset. Dug into it and here's why.
- when we call `setStore`, we have to pass in an object (ie, `{explanationPage: ''}`)
- this [object is run through `validateCookie` before it is allowed to be saved](https://github.com/cds-snc/ircc-rescheduler/blob/master/web/src/withProvider.js#L87)
  - this is how we make sure that all saved values are validated first
- validateCookie uses `fields` and `validate` from the current page and runs them against the object that was passed in.
- Since `explanationPage` was not part of [the Calendar's fields](https://github.com/cds-snc/ircc-rescheduler/blob/master/web/src/validation.js#L176), it was originally failing validation and then it wouldn't reset the value.
- By adding `explanationPage` in as a calendar field with the `'accept_anything'` validation rule, now the validation _will_ pass and the value will be reset.
- The side-effect of this is that our store will now include empty keys (keep reading)

**Note**
When validating fields on a page, any missing fields (ie, fields not passed in) will be reset to empty strings. 

This means that when we call `validateCookie({explanationPage: ''})`, it returns an object with all of the calendar fields: `{explanationPage: '', selectedDays: ''}`

Which means that we end up with an empty `selectedDays` key in the cookie under `explanation`, as you can see in the photo of the console output before and after validating the passed-in object.

console output 👇 
image

Similarly, the calendar page fields will now include an empty `explanationPage` key.  

Before:
```js
{
  store: {
    calendar: {selectedDays: [1,2,3]}
    explanation: {explanationPage: 'some string'}
  }
}
```

After:
```js
{
  store: {
    calendar: {selectedDays: [1,2,3] , explanationPage ''}
    explanation: {explanationPage: 'some string', selectedDays: ''}
  }
}
```

We don't need these extra keys but they don't actually affect our code at all because we never reference them, so they are pretty benign. 

The assumption that was broken was "pages should only ever update values that are on the page, not on other pages". 

There's probably a cleaner way to refactor this, but in terms of "least lines of code to fix the problem", this is the best.
</details>

timarney:
Okay.

I think I have issue 1 fixed.

Issue 2 thinking maybe something to do with validate cookie?

Console output of calValues:

screen shot 2018-09-25 at 9 33 05 am

timarney:
Nice work @pcraig3 that re-write seems to do the trick.

PR #476, opened 2018-09-25
NoJS - explanation page flow

timarney:
@pcraig3 thinking this needs the updated validate code from the other PR.


emanelfy:
**Background colour:** #eeeeee

**Placement:** At the top of the page underneath _"Citizenship appointments in October are scheduled on Tuesdays and Wednesdays."_

**Link (blue and underlined):** I cannot attend any of these appointments.

**Size:** Should match the checkbox text.

**Focus:** Even though this appears at the top of the page, the focus of this element should be the last thing before the submit button.

--
cc @timarney @pcraig3

pcraig3:
Okay, updated this to always reset the explanationPage value when calendar values are saved successfully.

To verify correct behaviour, you can add 
```
        <pre>{JSON.stringify(this.props.context.store, null, 2)}</pre>
```
to line 100 of the `ReviewPage.js` file.

web/src/withProvider.js
```diff
@@ -53,6 +53,11 @@ function withProvider(WrappedComponent) {
         if (val) {
           newCookie = setSSRCookie(res, key, val, prevCookie)
 
+          if (WrappedComponent.saveAfter) {
```
timarney: Nice - keeps things clear and at the Component level.

timarney:
Will approve this once it's passing (translations are in).

web/src/withProvider.js
pcraig3: yeah, it's nice that way. not nice in _every_ way, but you have to start somewhere.

timarney:
> Will approve this once it's passing (translations are in).

Looks like you have to approve it.  Original commit are mine.

PR #491, opened 2018-09-27
Final french language changes

anikbrazeau:


pcraig3:
This is not a failing PR, this was opened up before #492, which changed where our files are located.

PR #498, opened 2018-09-27
Remove respond by date

timarney:
Respond by date is no longer in use to calculate when the user will receive a response by.

The new setup is coded into the templates based on the family option.

### DIFF
--- a/src/withProvider.js
+++ b/src/withProvider.js
@@ -54,6 +54,7 @@ function withProvider(WrappedComponent) {
           newCookie = setSSRCookie(res, key, val, prevCookie)
 
           if (WrappedComponent.saveAfter) {
+            // eslint-disable-next-line security/detect-object-injection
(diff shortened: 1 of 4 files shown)


src/withProvider.js
```diff
@@ -99,8 +100,10 @@ function withProvider(WrappedComponent) {
         let newState = { [key]: obj }
 
         if (WrappedComponent.saveAfter) {
+          // eslint-disable-next-line security/detect-object-injection
           let { key: k, val: v } = WrappedComponent.saveAfter(newState[key])
           if (k && v) {
+            // eslint-disable-next-line security/detect-object-injection
```
pcraig3: whoops, good catch

pcraig3: not sure why this wasn't showing up before?

timarney: Ya not sure.

PR #501, opened 2018-09-28
Describe the service

timarney:
Add content to describe the service and it's purpose


pcraig3:
Note: I have a PR in development that adds more stuff to the README but this should go in as-is.

PR #506, opened 2018-09-28
Update README before handoff

pcraig3:
Added `Technical overview` section, table of contents, and emojis.

Probably going to clean this up some more, but I've been working on this for while and want to have it open as a PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,7 @@ We use several third-party services for easier development as well as tracking o
 - [Heroku](https://www.heroku.com/) watches our repository and builds an app for each pull request so that team members can easily verify correct behaviour of proposed changes (this feature is called [Heroku Review Apps](https://devcenter.heroku.com/articles/github-integration-review-apps))
 - [Snyk](https://snyk.io/) scans our [package.json](https://github.com/cds-snc/ircc-rescheduler/blob/master/package.json) file for packages with known vulnerabilities
 - [Sentry](https://sentry.io/for/react/) is used to capture JavaScript runtime exceptions in all environments (locally, on staging, and in production)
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -1,14 +1,59 @@
...
+
+Since we return server-rendered HTML and we have access to user data on the server, the Rescheduler works with or without client-side JavaScript, [serving users across the widest range of devices](https://digital.canada.ca/2018/08/08/supporting-users-gracefully-degrading-react/).
+
+The Rescheduler aims to make rescheduling an appointment as simple as possible. The technical implications of this are that we capture to smallest amount of data need to and we have minimal external dependencies.
+- User data is stored in browser cookies [which expire after a day](https://github.com/cds-snc/ircc-rescheduler/blob/master/src/cookies.js#L2)
+- There is no external database
+- Our one external API call is to [Amazon's Simple Email Service (SES)](https://aws.amazon.com/ses/) to send confirmation of requests to users and local immigration offices
```
timarney: Might want to mention NodeMailer - using SES as the transport layer.  Don't think we need to go into details but we're not using it direct.

If we do want to get into more detail ... because we needed attachments.

README.md
```diff
@@ -1,14 +1,59 @@
...
+
+### Use of third-party services 📮
+
+We use several third-party services for easier development as well as tracking our application out in the wild
+- [CircleCI](https://circleci.com/) runs automated tests on new pull requests and deploys new containers to the staging environment when new code is merged to the master branch
+- [Heroku](https://www.heroku.com/) watches our repository and builds an app for each pull request so that team members can easily verify correct behaviour of proposed changes (this feature is called [Heroku Review Apps](https://devcenter.heroku.com/articles/github-integration-review-apps))
+- [Snyk](https://snyk.io/) scans our [package.json](https://github.com/cds-snc/ircc-rescheduler/blob/master/package.json) file for packages with known vulnerabilities
+- [Sentry](https://sentry.io/for/react/) is used to capture JavaScript runtime exceptions in all environments (locally, on staging, and in production)
```
timarney: Just a thought - could mention something about source maps here + release hashes.  I don't think it needed but wanted to mention.

pcraig3: Yeah, that's a good point. Those things have been a bunch of trouble to get working so it makes sense that we at least mention them.

timarney: Oh and the stats / bundle size reports.

pcraig3: oh yeah 🤦‍♂️ 

too much tooling going on here

pcraig3:
Gonna admin merge this because we can always update later if there are grievous omissions or perilous misspellings.

PR #510, opened 2018-10-10
Privacy page fixes

timarney:


src/pages/PrivacyPage.js
```diff
@@ -38,9 +38,13 @@ const contentClass = css`
     display: block;
   }
 
-  pre {
+  .address {
     font-family: inherit;
   }
```
pcraig3: `<pre>` tags set their own font face, but `<div>`s don't, so we can get rid of this rule altogether.

src/pages/privacy/PRIVACY_fr.js
```diff
@@ -98,8 +101,6 @@ Vous pouvez obtenir plus d’information sur :
 - la façon dont nous vérifions que l’information que nous conservons est exacte et à jour
 - la façon de présenter une plainte
 
-For more information about the above issues, please contact the IRCC Access to Personal Information Coordinator through their <a href="https://www.canada.ca/en/immigration-refugees-citizenship/corporate/transparency/access-information-privacy.html" aria-label="IRCC Access to Personal Information Coordinator website">website</a> or by telephone at 1-800-667-6703.
```
pcraig3: Good save! ⚾️

PR #517, opened 2018-12-12
Add cloudformation infra scripts

dsamojlenko:


.cloudformation/production/cluster.yml
```diff
@@ -0,0 +1,513 @@
...
+    Properties:
+      RouteTableId: !Ref PrivateRouteTableTwo
+      SubnetId: !Ref PrivateSubnetTwo
+
+  # OPTIONAL: VPC Endpoint for DynamoDB
+  # If a container needs to access DynamoDB this allows a container in the private subnet
+  # to talk to DynamoDB directly without needing to go via the NAT gateway. This reduces
+  # the amount of bandwidth through the gateway, meaning that the gateway is free to serve
+  # your other traffic.
+  DynamoDBEndpoint:
```
obrien-j: I didn't think we were actually using any datastore in this case @dsamojlenko ...  Comment this out, or just pull it out. I'm good with either. It might be good as an example for future iterations.

.cloudformation/production/service.yml
```diff
@@ -0,0 +1,173 @@
...
+                 traffic to this service.
+  Priority:
+    Type: Number
+    Default: 1
+    Description: The priority for the routing rule added to the load balancer.
+                 This only applies if your have multiple services which have been
+                 assigned to different paths on the load balancer.
+  DesiredCount:
+    Type: Number
+    Default: 1
```
obrien-j: Any chance you could setup autoscaling, even just as an experiment. It's quitely likely this count will be sufficient for the load, even when fully rolled out to all offices, but wouldn't hurt.

If you could quantify costs of scaling up desiredcount, containercpu, and containermemory ,that'd also be helpful documentation to add to the project documentation

.cloudformation/production/service.yml
```diff
@@ -0,0 +1,173 @@
...
+          WorkingDirectory: '/web'
+          Memory: !Ref 'ContainerMemory'
+          Image: !Ref 'ImageUrl'
+          PortMappings:
+            - ContainerPort: !Ref 'ContainerPort'
+
+  # The service. The service is a resource which allows you to run multiple
+  # copies of a type of task, and gather up their logs and metrics, as well
+  # as monitor the number of running tasks and replace any that have crashed
+  Service:
```
obrien-j: A further extension to this would be zero-downtime deploys, using either a blue/green pattern or adding the newest deploy as an additional task definition within the service.

one example: https://github.com/aws-samples/ecs-canary-blue-green-deployment/tree/fargate
other: https://aws.amazon.com/blogs/devops/use-aws-codedeploy-to-implement-blue-green-deployments-for-aws-fargate-and-amazon-ecs/

PR #518, opened 2018-12-18
Nearly final round of updates

pcraig3:
Did another round of updates and went down a few rabbit holes. 🐰🐰🐰 

1. [Breaking change to emotion](https://github.com/emotion-js/emotion/blob/master/docs/migrating-to-emotion-10.md) meant that our federal banner disappeared without JS for some reason. Luckily, @nmakuch fixed it by recreating the Federal Banner in our app.
2. Since we were only using the ErrorBoundary from [GCUI](https://github.com/cds-snc/gcui) after Nick's change, I pulled it into our app as well and then removed GCUI as a dependency.
3. Some puppeteer tests started to fail randomly with some hard-to-track-down error, so I updated the "wait for this action to finish" time and swapped one set of tests over to query strings and that seems to have fixed it.

As of this point our packages are up to date, and we shaved a minuscule amount of code off our bundle size.


pcraig3:
gonna merge

PR #520, opened 2018-12-20
Don't show "Phone -" on 500 page if no location

pcraig3:
Our root domain has no location set, so we don't have contact information to show users.

Right now, it will say `Phone —` but then not have a phone number.

Instead, we should omit the Contact output entirely if there is no contact info.

## Screenshots

| before | after |
|--------|-------|
|  when there is a location ('vancouver' in the URL string), there is a phone number.  |  When there is *no* location, there is no phone number to call.   |
|  screen shot 2018-12-20 at 2 39 35 pm  |   screen shot 2018-12-20 at 2 39 42 pm  |

This PR removes the `Phone —` line in the second screenshot.

### DIFF
--- a/src/components/Contact.js
+++ b/src/components/Contact.js
@@ -107,4 +107,7 @@ Contact.propTypes = {
   showEmail: PropTypes.bool,
 }
 
-export { Contact as default, TelLink }
+const ContactWrapper = ({ ...props }) =>
+  getPhone() ? <Contact {...props} /> : null
+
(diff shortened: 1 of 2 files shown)


pcraig3:
I know! This is a great case for them! 😢

pcraig3:
Anyway, we can test on staging 😉

pcraig3:
Works now! 

https://rescheduler-dev.cds-snc.ca/500

PR #525, opened 2019-01-04
Update Razzle to fix nested package vulnerability

pcraig3:
This is a patch update that bumps Razzle in response to a Snyk vulnerability.

I've also added prop-types to `FocusedH1` because locally my linter started complaining.

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,7 +59,7 @@
     "nodemailer": "^4.7.0",
     "prop-types": "^15.6.2",
     "raven": "^2.6.4",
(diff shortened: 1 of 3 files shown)


obrien-j:
https://github.com/jaredpalmer/razzle/compare/v2.4.0...v2.4.1

Just putting this here for reference... Github makes is awkward to compare between tags.

pcraig3:
@timarney, yeah I added that in too because my linter was mad. But it's totally fine to do both imo.

PR #527, opened 2019-01-15
Disable GA

timarney:
This PR removes the Google analytics calls.  The functions have been left in place but the tracking calls are removed.


obrien-j:
Can we confirm a push to prod on this?

PR #528, opened 2019-01-17
Use POST requests in the no-JS flow

pcraig3:
Default form behaviour is to send `GET` requests back to the same URL except put all the keys+values in the URL as query parameters (eg, `/register?name=Paul&reason=travel`). Since this would mean that users' data would end up in the URL string, it would be possible for third-party services (like Google Analytics) to capture our users' personal information.

In order to obviate that, we are adding `method="post"` to all of our HTML forms and extending our code to accept `POST` request parameters instead of `GET` query variables. 

(The code was all lifted from the [`after-no-js`](https://github.com/pcraig3/after-no-js/compare/6cb500ccd4825e69ec59d4649e3285aa79c5c0d1...master) repo -- you can trial that at [https://nojs.now.sh](https://nojs.now.sh).)

Also updated one set of tests and deleted another set. More info in the commit messages if you're interested.

***

Note: using `GET` parameters will still work, but the app won't do this by default. If someone were to manually enter all their data in as url parameters, they would be accepted. Or, if they failed to validate, they would remain on the same page without any error messages appearing. I don't believe this is an issue.


pcraig3:
I don't now why @obrien-j didn't use github for this, but he gave the go-ahead on Slack:

> image
> John O'Brien

obrien-j:
Too many systems. A-ok by me!

PR #530, opened 2019-01-17
Add another sandbox flag to puppeteer

pcraig3:
These family check tests are pretty frequently causing errors in our deploy process, even through they run fine locally.

I've looked it up and there are a few issues on github where people add this extra argument when it runs on CI. Also I've split out the two family tests from each other. 

Maybe one of these will resolve our issue.

There's nothing wrong with the test, so I don't want to turn it off.


pcraig3:
Merging to get the app deployed to staging.

PR #533, opened 2019-02-06
Skip puppeteer tests on CI

pcraig3:
Something about our setup keeps failing these specific tests on CircleCI, so PR builds and releases keep failing.

There have been a couple attempts to ameliorate this in the past but these tests continue to fail intermittently.

Biting the bullet and removing them from CircleCI.

They can still be run locally with `yarn test:full`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -31,7 +31,7 @@ jobs:
 
       # run tests!
       - run: yarn a11y:test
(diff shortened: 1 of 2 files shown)


pcraig3:
Tim, I couldn't do it. 😢 

Do you want to have a pop, or should we just concede? 🏳️

PR #541, opened 2019-03-19
Add Content Security Policy and HSTS default headers

pcraig3:
Updated default headers with:

- a [Content Security Policy](https://helmetjs.github.io/docs/csp/)
- an [HSTS](https://helmetjs.github.io/docs/hsts/) header

And removed the `noCache` header as I don't think we need it.

Read more here:
- https://expressjs.com/en/advanced/best-practice-security.html#use-helmet
- https://helmetjs.github.io/docs/

### DIFF
--- a/src/server.js
+++ b/src/server.js
@@ -24,10 +24,8 @@ const server = express()
 const helmet = require('helmet')
 
 server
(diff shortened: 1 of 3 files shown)


obrien-j:
Looks good. The helmet magic is cool.

PR #544, opened 2019-04-16
Package bumps

timarney:
Package updates.

@pcraig3 note the Nodemailer was bumped (major) not sure what's the best way to test that.

This PR doesn't update to the latest React Router (had some failing tests on that).  Can look at that separately.


timarney:
Getting blocked on CI 

Your lockfile needs to be updated, but yarn was run with `--frozen-lockfile`

If we remove the --frozen-lockfile we run out of memory on the build.

